PR #14, opened 2017-11-09
fix missing include

metux:


carrotIndustries:
LGTM, but which compiler on what platform is failing due to this missing include?

metux:
devuan jessie
g++ (Debian 4.9.2-10) 4.9.2

in general, I'd recommend always explicitly including everything a compilation unit needs.
(not relying on recursive inclusion)

carrotIndustries:
If that's a gcc from 2014, this one's going to be the least of your problems...

metux:
On 09.11.2017 01:34, Lukas K. wrote:
> If that's a gcc from 2014, this one's going to be the least of your 
> problems...

I'm running devuan stable (actually, always running stable/lts releases
on production systems).

But right, there're more problems: gtkmm is too old (doesn't have GLArea
yet). For that particular case we'll yet have to sort out, how to handle
that new dependency - OpenGL. (eg. can we split it into multiple
packages, or do we need separate builds ?, ...)

I'd recommend add the constraint >=3.17.70 (least gtkmm version w/ 
GLArea) to the pkgconfig call.


--mtx

-- 
Enrico Weigelt, metux IT consult
Free software and Linux embedded engineering
info@metux.net -- +49-151-27565287

carrotIndustries:
In fact, the Gtk minimum version is 3.20 due to filechoosernative. By ifdefing all the filechoosernative things, the minimum gtk version should be 3.18. OpenGL 3 and GtkGLArea are a hard requirement for horizon since all rendering is done that way.

PR #24, opened 2017-12-09
Fix warnings

rnestler:
I started looking through the compiler output and fixing warnings where I considered the change to have almost no impact. I made separate commits for each warning I fixed, to make it easier for you to review and reject some of them if necessary.

### DIFF
--- a/util/accumulator.hpp
+++ b/util/accumulator.hpp
@@ -3,7 +3,9 @@
 namespace horizon {
 	template <typename T>class Accumulator {
 		public:
-			Accumulator(){};
+			Accumulator():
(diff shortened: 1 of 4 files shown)


util/accumulator.hpp
```diff
@@ -3,7 +3,9 @@
 namespace horizon {
 	template <typename T>class Accumulator {
 		public:
-			Accumulator(){};
+			Accumulator():
+				value()
+			{}
```
rnestler: I'm not sure about your preferred code style for member initializer lists, but that's what I'm used to.

PR #41, opened 2017-12-28
Add set-polygon parameter program command

fruchti:
I needed polygon shaped pads for an UFQFPN20 footprint.
2017-12-28-162742_800x900_scrot
I found no good way to enable parametrisation with parameter programs, so I added the `set-polygon` command which can set each vertex' coordinates individually.

`set-polygon [ <parameter class> <n_vertices> ]` takes 2 * n_vertices values from the stack (x and y alternating) and sets the polygon's vertex coordinates to those. Polygons with arcs or with any other number of vertices than n_vertices currently throw an error.

### DIFF
--- a/pool/padstack.cpp
+++ b/pool/padstack.cpp
@@ -85,6 +85,37 @@ namespace horizon {
 		return {false, ""};
 	}
 
+	std::pair<bool, std::string> Padstack::MyParameterProgram::set_polygon(const ParameterProgram::TokenCommand *cmd, std::deque<int64_t> &stack) {
+		if(cmd->arguments.size()<2 || cmd->arguments.at(0)->type != ParameterProgram::Token::Type::STR || cmd->arguments.at(1)->type != ParameterProgram::Token::Type::INT)
+			return {true, "not enough arguments"};
+
+		auto pclass = dynamic_cast<ParameterProgram::TokenString*>(cmd->arguments.at(0).get())->string;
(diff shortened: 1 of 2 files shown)


carrotIndustries:
Looks good to me, but do we really need this, or can we just leverage the `polygon-expand` command from the packages? With the latter command, you could easily expand/shrink a polygon.

Or do you want to make the size of the chamfer adjustable by a parameter? If you're aiming for that one, we do need this command.

fruchti:
As far as I see, `expand-polygon` allows only one parameter from the stack. It would be therefore impossible to adjust width and height of the pad independently.

To be able to adjust the bevel would be really nice, but that would also need custom parameters. In my example, with a adjustable bevels I would only need one padstack for the mirrored variants of the pad.

fruchti:
Is there anything holding this back? I think `set-polygon` would be the general way for handling a lot of cases which aren't possible right now.

The only thing I'm unsure about is how arcs should be handled (if they had to be included).

carrotIndustries:
The only thing holding this back is the code review above. By simply clearing all vertices, we could also get rid of the intermediate array.

fruchti:
Sure thing. Yet now, it has to iterate over all polygons for each vertex. Is that really better?

carrotIndustries:
I squash merged your branch with some minute changes (replaced push by emplace) in https://github.com/carrotIndustries/horizon/commit/e286542fd1df93851a9af6d550b971957447c682

PR #59, opened 2018-01-10
cleanup of project sources

peterus:
- moved project-source files to 'src'.
- moved 3part-source files to '3part'.
- changed Makefile to get everything compiled again.
- some other small changes:
  - moved export_pdf.hpp/cpp to src/export_pdf/export_pdf.hpp/cpp
  - imp.gresource.xml (remove whitespaces)
  - src/canvas3d/canvas3d.cpp, src/imp/imp_schematic.cpp (include directory update)


carrotIndustries:
LGTM overall, some things I'd like to see addressed before merging:
 - 3part isn't a particularly good name, "lib" or "3rd_party" would be much better
 - router/pns_horizon_iface.cpp is mostly my code, so it's better off in src

peterus:
changed according to your suggestions :)

PR #74, opened 2018-01-23
Added style reference (.clang-format), style check scripts and CI checks

korken89:
For people using editors, any clang-format integration should detect the .clang-format and work as clang-format itself does the search of directories to find it.


carrotIndustries:
Do we really need the 80 chars line length limit? https://github.com/korken89/horizon/blob/master/src/imp/imp_package.cpp#L357 isn't really pretty.  120 .. 150 leaves this line as one.
I also prefer the `else` being on its own line, i.e. not on the same line as the `}` of the preceding if.

On Arch Linux clang-format is available as `clang-format`, not `clang-format-4.0`, no what it's like on other distros.

korken89:
I have updated as you requested.
I personally think that one should never use it with more than 80, as it works for everyone developing.
But I also have to reason to change the way you work and more and more like to have more than 80 columns. :)

Have a look now, tell me what you think!

korken89:
When it comes to clang-format-4.0, it is the most widely used version of clang-format that is available.
In arch it should be available as well, but generally it is aliased as only clang-format.
The issue is that different versions of clang-format can give slightly different results, which is a pain to handle in the CI checks.

korken89:
Okey I had more look into Arch @carrotIndustries .
Since Arch / Manjaro are using a rolling release model there will not be something like `clang-format-X.0`.

We could look into making the script smarter than it is now, to detect version of clang-format, if it is 5.0, use it - else look for clang-format-5.0, and if that does not exist let the script fail with error message.
What would you want?

carrotIndustries:
Thanks for your effort and for integrating my specific requirements!

korken89:
No problem!
I was about to start using KiCAD but was shown this by a friend (who is a developer in KiCAD), and I really hope this will become something great!
The only thing problematic is that it is not supported in Ubuntu 16.04, but this can be fixed.

carrotIndustries:
Yeah, IIRC the only blocker for Ubuntu 16.04 is Gtk 3.18 not supporting GtkFileChooserNative (apart from the odd compiler dumbness). Should be easy to sort out using a bunch of ifdefs or a wrapper. OTOH, the next LTS is just a few months away, that's why Gtk 3.18 compatibility didn't seem worth it to me.

PR #80, opened 2018-01-30
Use "previous" icon instead of close button, separate buttons to prevent confusion

jeanthom:
I find the "Close" button inappropriate in the Pool Manager, I think that a "Previous" button would be more appropriate.

I also used different packing position to separate "Previous" action and "Update pool" action.

### DIFF
--- a/src/pool-mgr/window.ui
+++ b/src/pool-mgr/window.ui
@@ -474,19 +474,19 @@
             <property name="visible">True</property>
             <property name="can_focus">True</property>
             <property name="receives_default">True</property>
-	    <property name="tooltip-text" translatable="yes">Return to pool list</property>
+            <property name="tooltip-text" translatable="yes">Return to pool list</property>
 
-	    <child>
-	      <object class="GtkImage" id="navigation_previous_icon">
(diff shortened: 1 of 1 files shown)


carrotIndustries:
Replacing the "Close" button with a "back" one is definitely the right thing to do. I'll still have to make up my mind about the placement of the update button, but putting it to the left side probably fine too.

While you're at it, have a look at the project manager as well, the UI is pretty similar.

carrotIndustries:
> but putting it to the left side probably fine too.

Oops I meant the right side. TBH, my judgment in that regard is rather limited since I stared at that window for way too long and initial placement went without much thought. What looks best to you?

jeanthom:
I think seperating action buttons from navigation buttons when possible is a good idea, however this is a personnal preference, and I'm not a UX expert

carrotIndustries:
I had a look at various gnome applications, they all have the non-navigation buttons on the right side, so putting the "update pool" button on the right side is the way to go.

jeanthom:
Maybe we should merge this PR and then tacke interface uniformisation elsewhere ?

PR #84, opened 2018-01-31
Reorganize pool manager's remote tab layout

jeanthom:
For comparison :
Before
After

The current layout wastes a lot of space, and doesn't fit on small screens (like 1366x768, which is the most used resolution according to those statistics : https://www.rapidtables.com/web/dev/screen-resolution-statistics.html).

The github link on the top left corner has some padding that can only be disabled using CSS. As I'm not familiar enough with Horizon's source code, I won't fix it for now.


carrotIndustries:
Looks like a was a bit too concerned about wasting vertical space when designing that tab...
This is how it looks since the commit reference above:
2018-01-31-205838_983x605_scrot

jeanthom:
Even nicer !

PR #91, opened 2018-01-31
cmake build

peterus:
There are still two issues:

Linux:
- horizon-prj-mgr -> "Top Schematic" and "Board" Icons are missing

Windows:
- horizon-*-mgr -> resources are not loading

As I was already looking into this issues for some hours, maybe somebody else should take some minutes to check the problems.


jeanthom:
Build is not working on my computer, I get "No such file or directory" about `BRepBuilderAPI_MakeWire.hxx` in `export_step.cpp`.

Also do not hardcode "-j" flag in Make, as this parameters depends on the computer you're using.

peterus:
@jeanthom :
you have to install opencascade, than it should be found.
in my cmake files are no "-j" parameters

peterus:
i found out, that on windows (even with msys2) the cmake package to handle the  is adding a "C:" to the prefix:

`<gresource prefix="C:/msys64/net/carrotIndustries/horizon">`

of course it is not like this in the code, so it can not find the resources.
I will try to fix the cmake-package.

I will also revert some changes which are currently just too much to handle (moving the canvas files somewhere else etc.)

jeanthom:
I have OCE-devel installed on Fedora 25 x64, and normal build works OK.

"-j" parameter is hardcoded to 10 in `./compile.sh`

carrotIndustries:
Great work so far, however I'd like to have this as several commits for easier review and meaningful history:
 - Remove unnecessary files
 - (Reorganize files)
 - Switch to cmake

To me, the new directory structure seems rather arbitrary to me, in particular I don't like the shaders not being close to the related source files anymore.

peterus:
@jeanthom ah yes, thats true! will remove the file and move the calls to the different CI files

@carrotIndustries 
my intention was to separate all resources from the code. but i changed already my mind and will leave it as it was.
good point with the commits, also easier for me to deliver the code.

-> will close the PR and create a open one with the changes pointed out from you.

PR #95, opened 2018-02-05
Remove default size for fabrication output window

jeanthom:
The default window size doesn't fit on small screens. Removing the default width/height attributes fixes that.


carrotIndustries:
Not sure about merging this: Users with large screens may complain that the dialog is showing unnecessary scroll bars. What effect has setting "propagate natural height" on the scrolledwindow in your case?

jeanthom:
It seems to work great on my computer.

Issue #89, opened 2018-01-31
Externalize CSS code into separate files


jeanthom:
Currently CSS code is embedded in C++ source code, which is not practical (terrible CSS syntax highlighting, etc.). We could use `Gtk::CssProvider::load_from_file` and making it useable from the whole application.

carrotIndustries:
Instead of using `load_from_file`, we should use `load_from_resource` so the binaries stay self-contained and don't have to go searching for files at runtime.

jeanthom:
You're right. CSSProviders are used here :
 * [ ] src/imp/imp.cpp
 * [ ] src/imp/3d_view.cpp

PR #96, opened 2018-02-05
Moving CSS code outside C++ source

jeanthom:
Related to issue #89


carrotIndustries:
Doing this once in the imp initialization code should be enough, since the CSS is common to the whole application.

jeanthom:
I thought it wasn't the case as all those classes represent different windows. I removed the CssProvider code and asserted that my changes wouldn't break anything.

PR #99, opened 2018-02-07
Fix permissions in clang format script and bugs in run_clangformat.sh

jeanthom:
I noticed that the permissions weren't right (not executable) and also that the if conditions in run_clangformat.sh were not written correctly.


jeanthom:
Style checks fail because current Horizon source code doesn't meet clang-format requirements. Should we merge this and then correct code format in another pull request, or should we use that pull request to apply recommended code formatting ?

PR #103, opened 2018-02-08
New selection filter, using .ui files for interface description

jeanthom:
Here is the new design :
capture d ecran_2018-02-08_01-11-45

What disturbed me in the previous design : the "select all" button icon wasn't explicit, the arrow right in front of each checkbox means to me that there was something to expand but no.
Regression : you cannot select ONE item anymore by clicking on a button. If that feature is really important, we could reimplement it.

### DIFF
--- a/imp.gresource.xml
+++ b/imp.gresource.xml
@@ -42,6 +42,7 @@
     <file>imp/fab_output.ui</file>
     <file>imp/3d_view.ui</file>
     <file>imp/step_export.ui</file>
+    <file>imp/selection_filter.ui</file>
     <file>prj-mgr/part_browser/part_browser.ui</file>
 
     <file>global.css</file>
--- a/src/global.css
+++ b/src/global.css
@@ -27,8 +27,3 @@
 .bus-toolbar {
   border-right: 1px solid @borders;
 }
-
-button.imp-selection-filter {
-  font-size: 11px;
-  padding: 3px;
-}
--- a/src/global.css
+++ b/src/global.css
@@ -27,3 +27,8 @@
 .bus-toolbar {
   border-right: 1px solid @borders;
 }
+
+button.imp-selection-filter {
+  font-size: 11px;
+  padding: 3px;
+}
--- a/src/imp/imp.cpp
+++ b/src/imp/imp.cpp
@@ -219,8 +219,8 @@ void ImpBase::run(int argc, char *argv[])
     warnings_box->signal_selected().connect(sigc::mem_fun(this, &ImpBase::handle_warning_selected));
     main_window->left_panel->pack_end(*warnings_box, false, false, 0);
 
-    selection_filter_dialog =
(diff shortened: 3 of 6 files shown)


carrotIndustries:
> you cannot select ONE item anymore by clicking on a button. If that feature is really important, we could reimplement it.

That feature is really important to at least me, since selecting only one object type is a common use case.

Some more nitpicking:
 - Don't override font sizes and padding unless really needed
 - This dialog doesn't use header bars whereas every other dialog does
 - The listbox doesn't expand with the window
 - Your layout has has very little padding, doesn't look consistent with the rest of the application

Overall, only these aspects of the current implementation may need improvement:
 - Find another icon for the "select one" button
 - Move the "select all" button from the header bar to the main window

jeanthom:
peek 08-02-2018 13-07

I can't get the window size to be wide enough so that the title is displayed correctly :frowning_face:

jeanthom:
capture d ecran_2018-02-08_14-02-13

carrotIndustries:
I'm not quite sure whether differentiating between the check mark and its label is a good idea from a usability POV, we should probably retain that extra button.

The padding thing wasn't referring to the padding around the scrolledwindow, but rather to the individual elemement's padding.

Having that large button in the doesn't look particularly pretty, so either find a more suitable icon or place the button inside the actual window.

jeanthom:
> I'm not quite sure whether differentiating between the check mark and its label is a good idea from a usability POV, we should probably retain that extra button.

I agree on that point, but I can't find an icon that carries the right message. The current "select only" icon is more suited for folded content.

Also, is the selection filter really used that much to justify a button in the header bar ? Why not integrate the selection filter into the GtkPopover for selection settings ?

atoav:
When doing multiple footprints for TSSOP yesterday I used this thing a lot in its old state. What I thought from a usability standpoint:  
- I think the whole thing could be in the left sidebar instead of a floating window. Typically this kind of modal window style used in Blender for example) helps to use screen space more effectively.
- I'd generally like to have the ability to click drag over multiple checkboxes (this would be great for all toggleable lists in horizon, like layer visibility etc). See the way photoshop did it:  
Photoshop layer selection

jeanthom:
The click'n'drag won't be easy to implement, but it seems like the right thing to do.

atoav:
@jeanthom I thought that it wouldn't be easy but it could be re-used more than once. Maybe one could ask the [guys at Blender](https://wiki.blender.org/index.php/Community:Chat) for pointers, because they have the same implemented in their Outliner:   
blender_drag_selection

jeanthom:
Blender has their own interface toolkit, so we can't use their code unfortunately :-(

carrotIndustries:
0ec5457883bf7bf25868ba0b4f0d21e3572d5573 does away with the arrow and changes the button for selecting all object types

PR #107, opened 2018-02-08
Ask for text before entering text tool

jeanthom:
I guess nobody feels the need for placing "TEXT" labels in his/her schematic sheet :-)


carrotIndustries:
LGTM, but do we really need the extra text variable? Just save the pointer to the old text before creating the new one.

PR #109, opened 2018-02-09
Replace new/open project button labels with icons

jeanthom:
capture d ecran_2018-02-09_14-01-50

Tooltips were added to the buttons


carrotIndustries:
Text is just fine, space isn't an issue. The pool/project manager's start page more or less a copy of gnome builder, let's keep it that way. https://blogs.gnome.org/chergert/files/2017/03/Screenshot-from-2017-03-19-23-15-41.png

PR #111, opened 2018-02-11
Switch to cmake

peterus:
Here are finaly the changes for cmake!
there are 10 commits, if you look into the comments you will see what i have changed in this commit.

the PR is working now on Linux (Arch) and Windows (Win10) (except CI tests which have to run now).


peterus:
Looks like there are still some issues (Include path of "BRepBuilderAPI_MakeWire.hxx" and on Windows it is not generating the correct makefiles. Will continue the next days.

peterus:
finally all CI issues fixed.

carrotIndustries:
Great work!

Some things ended in the lib folder that shouldn't be there:
 - All the imp stuff. That should be in src/imp
 - The property_panels would be best off in src/imp/property_panels

Would you mind opening a new PR with the CI fixes integrated into the previous commits?

peterus:
regarding the imp stuff:
i was trying it as you have explained it (clean files/code) BUT unfortunately there are some include dependencies which had to force me to put the files there (i was checking every file with the compiler and ended up with this setup).
As i don't want to put here a code refactoring i was taking this approach to open the PR as it is. Next step for a new PR would be to refactor the imp files and code.

I would like to suggest to use this PR now as it is, because i spend already many hours to have it this far. the next problem is: if there are new and difficult changes coming in, i have to start again from the beginning (this is already my 4 attempt).

CarloMara:
@peterus build goes well on ubuntu. No more missing BRepBuilderAPI_MakeWire.hxx

thanks,
Carlo

peterus:
@CarloMara can you check the pool-mgr when downloading a new pool? 
#112 is looking like there is still somewhere a problem. I can not check it at work, because we have very strange proxy settings and i am always getting a cert error.

carrotIndustries:
Without actually checking: Looks like you forgot to add the sql schemas to the glib resources.

peterus:
this sql-files are used for the resources:
- lib/util/window_state_schema.sql
- lib/pool-update/schema.sql
- pool-update-parametric/schema.sql

carrotIndustries:
> i was trying it as you have explained it (clean files/code) BUT unfortunately there are some include dependencies which had to force me to put the files there (i was checking every file with the compiler and ended up with this setup).

That's odd, at least the imp_schematic.cpp has to move from lib to imp to maintain consistency.

peterus:
I was just taking a look into the source code: The Interface class has some strong dependencies to ImpSchematic (imp_schematic.cpp). Here is definitely a break in the Interface concept which has to be fixed!
Normally the Interface class should be derived and not set as friendly, than this can not be happening.

carrotIndustries:
The ImpInterface isn't an interface in the typical sense, it's used by the tools to interact with the imp (show dialogs, show toolbar, etc) The reason for implementing the interface as a separate class was to clearly define the interaction between tools and imp.

So why exactly are header dependencies an issue with cmake?

peterus:
the issue is not with cmake. i just want to define a clean area between code which is just running inside of a program (like 'pool-mgr') and common stuff ('lib' folder).
You want to move now imp_schematic.cpp from the 'lib/imp' to 'imp'. But this is not working as it is a dependency from the Interface class (against common design rules, how to get an abstraction inside the code)

carrotIndustries:
In that regard,

lib/imp
lib/core
lib/property_panels

should move to imp since that's the only place where these are used.

CarloMara:
hey @peterus,

I'm having some problems when I compile via Clions. At the linking stage it fails complaining about some missing shared objects from glib, here is the error:
```
/usr/bin/ld: ../../../lib/libextern_dxflib.a(dl_writer_ascii.cpp.o): undefined reference to symbol '_ZNKSt9basic_iosIcSt11char_traitsIcEE4failEv@@GLIBCXX_3.4'
//usr/lib/x86_64-linux-gnu/libstdc++.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
src/imp/CMakeFiles/horizon-imp.dir/build.make:621: recipe for target '../bin/horizon-imp' failed
make[3]: *** [../bin/horizon-imp] Error 1
CMakeFiles/Makefile2:288: recipe for target 'src/imp/CMakeFiles/horizon-imp.dir/all' failed
make[2]: *** [src/imp/CMakeFiles/horizon-imp.dir/all] Error 2
CMakeFiles/Makefile2:300: recipe for target 'src/imp/CMakeFiles/horizon-imp.dir/rule' failed
make[1]: *** [src/imp/CMakeFiles/horizon-imp.dir/rule] Error 2
Makefile:157: recipe for target 'horizon-imp' failed
make: *** [horizon-imp] Error 2
```
I don't really know what to do, because if launch the building manually it works like a charm. I've already checked the usual suspects, like building using different a tool-chain but no luck so far.
  
Attached you find a more verbose trace, produced with `-v` on gcc. 

Carlo

[linking_failed.txt](https://github.com/carrotIndustries/horizon/files/1768231/linking_failed.txt)

carrotIndustries:
> What is missing to get this merged?

The main blocker is that I'm not quite happy with the files being reorganized the way it's done in this PR.

Someone also needs to remind me of the advantages of cmake over plain make. After subtracting the lines of source definitions the current Makefile is less than 100 lines, the CMakeLists.txt amount to more than 200 after subtracting the resource list.

To me at least 17 files totaling >200 lines on how it's put together is less approachable than a single 100 line Makefile.

CarloMara:
The main advantage for me is to be able to use clion, which IMHO is far superior to any other C/C++ ide.

Cheers,
Carlo

peterus:
As the project owner is not willing to continue with a project setup which has a a lot of benefits like:
- getting a good structure in code and files (you have currently a big problem in you structure, currently it will run but in the future you will spend a lot of hours fixing this),
- be able to run the compilation on many different platforms without taking care of paths or other setup stuff
- and having a package checking.

and the main reason: cmake IS today's standard! Makefile is not the way to go anymore in an c++ world.

all this is leading me to this conclusion: i will not continue to drive the cmake-switch. I spent already to much hours for nothing! Thanks!

greenscreenflicker:
I understand @peterus feelings, i share the same: If you help and your work is not valued by the project owner, it's very frustrating. I had the same issue with submitting parts to the pool. @carrotIndustries I think the project has potential. Let the community help and encourage them to contribute.

peterus:
last week I was at the "We are developers" conference in Vienna. I listen to many talks and specially to open source talks. One thing what was highlighted the most was that we are not working for/with the source code, we are working with different kind of people. And as a project owner you have to look how you communicate with your contributors. specially if you don't want an pull request.

I am already since many years a C++/C developer for embedded systems (not really visible on my github account as I do not have so much time to share my work).
I have a lot of knowledge how to structure source code, how to setup a good and working workflow, style checking, linter and many more. I am maybe not the best for GUI or other things, but I love to refactor and clean the code to make the source code better and faster.

I really like this project and I am looking forward how it will develop in the future (I still have the project on my "watch"). I also would like to add my support in developing updates and creating pull request. But as I never know if my changes are accepted, I am currently not really willing to give here my support. Specially if there are already new updates prepared which are based on my pull request.

pwnorbitals:
It's been two years now, and I believe it is the concensus for c++ developpers and project to make the switch to CMake (or better, Meson). Is there any plan to accept this PR or bring constructive criticism to fix the current PR @carrotIndustries ? Otherwise, we should just close it and move on

carrotIndustries:
As written before, the cmake files proposed in this PR seems more complex than the current Makefile. 

>  switch to CMake (or better, Meson)

As meson's used by a lot of projects also using gtk and its syntax is nicer than cmake, I think that meson's the best option to switch to.

PR #177, opened 2018-07-15
Pool manager: take package graphical elements into account for default zoom level

endofexclusive:
This patch set changes the way default zoom level is calculated in the pool manager. For packages, it tries to take graphical elements into account in addition to pads.


carrotIndustries:
d2e52d2a37ab704fb97258af8d68a0804176d643 fixes this by using vertex data from canvas to fit view

PR #181, opened 2018-07-26
Support build against OCCT

kkremitzki:
Hi, in Debian (and eventually Ubuntu), liboce (OpenCASCADE Community Edition, a fork which has stalled) is on its way to being replaced by the original, OCCT (OpenCASCADE Technology) or libocct. I saw that there's a package for this software in Debian by searching for reverse-dependencies since the plan is to eventually remove liboce.

This patch seems to be enough to successfully build, but I didn't do any further testing. Currently in Debian Testing we have OCCT 7.3 (2018), whereas liboce is based on OCCT 6.9 (2015), so there is a possibility of bugs coming down the pipeline, but since I'm not familiar with this software or how it uses OCCT I can't probe much further.


carrotIndustries:
we can now override the `INC_OCE` variable in the makefile to address this

PR #183, opened 2018-07-31
Do not depend on $(HORIZON_POOL)/tmp

endofexclusive:
Pool manager *Part Wizard* crashes if $POOLDIR/tmp does not exist. A
workaround is to make sure the $POOLDIR/tmp always exists.

This commit fixes the crash and dependency by asking GLib for a
temporary directory with get_tmp_dir().


endofexclusive:
Thank you for merging.

With this patch in place, the `mkdir_pool(base_path, "tmp")` in `view_create_pool.cpp` may not be needed any longer.

PR #184, opened 2018-07-31
Added support for pin direction in the part wizard pin importer.

endofexclusive:


carrotIndustries:
I've had something similar on my mind for quite some time, but thought about providing UI for mapping the direction values provided by the user to the horizon ones.

I'm not quite happy with your implementation making the UI text part of some sort of API contract, a better solution (if we're too lazy for aforementioned UI) would be to make the `pin_direction_lut` public and use the names provided in there.

endofexclusive:
I agree about the UI text dependency.

A broader approach would be to allow the user to import from a CSV file directly. CSV is much closer to the user domain than JSON, can be exported directly from spreadsheet programs, etc.

The CSV format could be something like:
`padname,pinnames,direction,gate`

Alt. pinnames could be space separated in `pinnames` after the primary one. Another option is to add alt. names, comma separated, after the `gate` field.

carrotIndustries:
From a user's perspective CSV would definitely be nice. But parsing user-supplied CSV is harder to get right and users (might) expect UI in order to import their particular flavour of CSV. So I went with JSON since that's more well specified and I already had a parser at hand.

carrotIndustries:
fixed in 48337827c68e5a950cbefb2469ce79f7e7ae3df0

PR #191, opened 2018-08-07
Power symbol: support more orientations

endofexclusive:


carrotIndustries:
Great work! Nitpicking, mirroring horizontal power symbols should change their orientation to make things consistent.

endofexclusive:
3a18ca31da8557fdab002e6745ec249d400b2f14 should do it

After playing around with this, it seems it would be useful to replace the `mirror` boolean with another `Orientation` which describes how the text is placed relative to the symbol. Then the net name could be located above or below the symbol as well. That is kind of typical for a schematic GND symbol.

For example:
```
VCC O
    |


   VCC
    O
    |


    O VCC
    |


     VCC
    -O


    -O VCC


    -O
     VCC


    |
    O VCC


    |
    O
   VCC
```
etc.

That is, 12 different combinations in total for rotation.
This would not be fully compatible with how `move_mirror_or_rotate()` is called in `ToolMove::begin()`. It assumes that MIRROR_Y is equivalent with two rotations. One solution to that could be to split `move_mirror_or_rotate()` into one mirror and one rotate functions.

carrotIndustries:
> It assumes that MIRROR_Y is equivalent with two rotations. 
Equivalent to mirroring around X axis and rotating by 180°.

> After playing around with this, it seems it would be useful to replace the mirror boolean with another Orientation which describes how the text is placed relative to the symbol. Then the net name could be located above or below the symbol as well. 

Seems reasonable, any idea on how to to name the various placements? Left/right is ambiguous since the actual placement will depend on the power symbol's orientation.

PR #197, opened 2018-10-12
Fix spacing calculation in DIL footprint generator

fruchti:
The calculation of the horizontal pad spacing wasn't consistent with the drawing.


carrotIndustries:
Thanks for spotting and fixing!

PR #208, opened 2018-11-25
initial net label size fixup

endofexclusive:
Before: "net_label_size" defaulted to 0 mm and stored in
DRAW_LINE.json.

Now: defaults to 1.5 mm and stored in DRAW_NET.json.


carrotIndustries:
duh, thanks for spotting and fixing

PR #209, opened 2018-11-25
Fix warnings

endofexclusive:
Removes warnings as reported by Clang 6.0.0.

Commits are ordered by type of action, and with 3rd party in separate commits.

```
$ c++ -v
FreeBSD clang version 6.0.0 (tags/RELEASE_600/final 326565) (based on LLVM 6.0.0)
```

The directory `3rd_party/dxflib/` is not fully covered by this patch series. It still has some `-Wshadow` warnings.

### DIFF
--- a/3rd_party/clipper/clipper.cpp
+++ b/3rd_party/clipper/clipper.cpp
@@ -3200,7 +3200,7 @@ void Clipper::BuildResult(Paths &polys)
     int cnt = PointCount(p);
     if (cnt < 2) continue;
     pg.reserve(cnt);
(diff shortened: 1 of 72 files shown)


endofexclusive:
Yes cherry-picking the Horizon specific commits is the way to go.

If reaching low number of warnings is a goal, then it could be investigated to use per-directory specific warning flags.

carrotIndustries:
(almost) all warnings resulting from actual application code are fixed in deee6a58c8f82b39befe175110177abc6dc7294c

PR #227, opened 2019-01-26
Added Layer descriptions

atoav:
Added all layer descriptions. Most are based on the texts used in [the wiki](https://github.com/carrotIndustries/horizon/wiki/Creating-a-Package), but I took some freedom to shorten them a bit or make them more concise.  

I am not sure whether I forgot to include some rules or requirements, maybe @fruchti should also have a look on these.

Another thing: it might look better to vertically align the text to the top, not sure how to do that myself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ dist
 docs
 */__pycache__
 src/preferences/color_presets.json
+src/imp/layer_help.json
--- a/Makefile
+++ b/Makefile
@@ -342,6 +342,7 @@ SRC_IMP = \
 	src/export_bom/export_bom.cpp\
 	src/widgets/unplaced_box.cpp\
 	src/widgets/tag_entry.cpp\
+	src/widgets/layer_help_box.cpp\
 
 SRC_ROUTER = \
 	3rd_party/router/router/pns_router.cpp \
@@ -535,6 +536,9 @@ endif
 
 src/preferences/color_presets.json: $(wildcard src/preferences/color_presets/*)
 	python3 scripts/make_color_presets.py $^ > $@
+
+src/imp/layer_help.json: $(wildcard src/imp/layer_help/*)
+	python3 scripts/make_layer_help.py $^ > $@
 
 src/resources.cpp: imp.gresource.xml $(shell $(GLIB_COMPILE_RESOURCES) --generate-dependencies imp.gresource.xml |  while read line; do echo "src/$$line"; done)
 	$(GLIB_COMPILE_RESOURCES) imp.gresource.xml --target=$@ --sourcedir=src --generate-source
--- a/imp.gresource.xml
+++ b/imp.gresource.xml
@@ -101,5 +101,6 @@
     <file>canvas3d/shaders/background-fragment.glsl</file>
 
     <file>preferences/color_presets.json</file>
+    <file>imp/layer_help.json</file>
   </gresource>
 </gresources>
--- a/scripts/make_layer_help.py
+++ b/scripts/make_layer_help.py
@@ -1,0 +1,14 @@
+#!/usr/bin/python
+from __future__ import print_function
+
+import json
+import sys
+import os
+
+out = {}
+for filename in sys.argv[1:] :
+	txt = open(filename, "r").read().strip()
+	layer = os.path.basename(filename)
+	out[layer] = txt
+
+print(json.dumps(out, sort_keys=True, indent=4))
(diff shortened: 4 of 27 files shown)


carrotIndustries:
> Another thing: it might look better to vertically align the text to the top, not sure how to do that myself..

Fixed that.

>The lines are very long in the text files.

Since the text gets displayed as-is, that's how it's supposed to be. Any half-decent editor will wrap the lines to a more acceptable length.

>Can they be named LAYER_INNER_COPPER, etc instead? Preferrable with matching enums
in the implementation instead of magic numbers?

Right now, there's to mapping to strings that we could make use of. Adding a mapping just for this purpose isn't really worth it.

atoav:
This reaction is longer than demanded by this topic, because it seems to touch some areas that need discussion.

## pool agnosticism
> I fully support the approach to describe what the layer means and how the program uses the information. I would however recommend to not include andy design rules (dimensions etc) in the descriptions.

I agree that horizon should be somewhat "pool agnostic" in the sense, that developement should be made without too much assumptions about which pool is going to be used. This means it should always be possible to do whatever is possible without having *enforced* rules about how things have to be.

That beeing said I don't really think the tips *force* anybody to do anything in a particular way. Anyone who maintains a pool is free to ignore these hints (in fact I do so myself), but it is a good starting point for new users. 

## Pros/Cons of including design rules
What speaks for having these numbers in general is:
- It helps new and inexperienced users to make choices that work as a sensible default¹
- Other than enforced rules it doesn't force anybody to do anything in a certain way
- It helps experienced users to remember what the public pool uses, should they decide to contribute parts
- As a nice side effect it reduces the work load for those who manage pull requests into `horizon-pool`

What speaks against it is:
- Professional users probably don't need that kind of reminder
- It is unclear what this means for the long term development of the pool ecosystem (there have been no real discussions about how this should look like in the long term)

Maybe we should change the wording to make it sound more like a recommendation to new users? More a *can* than a *must*?

## What is the scope?
Apparently this raises questions about the scope horizon has as a open source project – goals, non-goals, target groups – as well as how the pool ecosystem should be structured in the future (I mentioned some thoughts on this (here)[https://github.com/carrotIndustries/horizon/issues/178#issuecomment-455857456]).

IMO horizon should be easy to start with for non-professionals, while providing steamlined workflows for more professional users. Reducing the friction for both sides is a good goal (and contrary to popular opinion they are not mutually exclusive). If these targets are taken seriously, the public pool needs to be an easy default for new users: easy to get, easy to update, easy to contribute to and easy to move away from.

At the same time I'd love the idea of a more "federated" or decentralized system of pools, which all can be used at the same time, without to much hassle.

Any opinions on this?


-----
¹ A space of ≥0.2mm between package outline and silkscreen are a good idea from a purely rational standpoint, because everything else will cover the silkscreen. The 0.15mm line width for the silkscreen is a good guideline, because most PCB manufacturers I know use 0.15mm as a recommended minimum line width for their screen printing process. So going lower might not be a good idea, while going wider than that doesn't make much sense in terms of readability.

fruchti:
I like it! One minor nitpick: SI mandates a space between number and unit, so it should be “0.2 mm” instead of “0.2mm”, etc (preferably with a non-breaking space).

I agree that Horizon should be pool-agnostic, yet that does not mean there can't be any default rules in these help texts. It should be made clear that these rules are for the default pool and not limitations of Horizon EDA, though. I'd suggest to add a summary of each layer's most important rules as a last paragraph to the help texts, with a separate heading, for example “Default pool recommendations”.

carrotIndustries:
Instead of making these texts part of horizon itself, how about loading these texts from the pool? I'm thinking of a directory `layer_help` at the root of the pool containing the files that are now in `src/imp/layer_help`.

atoav:
> Instead of making these texts part of horizon itself, how about loading these texts from the pool? I'm thinking of a directory layer_help at the root of the pool containing the files that are now in src/imp/layer_help.

That is even better! It allows professional users to add their own hints there, while everybody who uses the public pool gets these hints. I admittedly feel a little silly for not having had that idea.

fruchti:
I like that idea! I guess these could be inherited from other pools or overwritten, but how would pools inheriting from multiple other pools choose them?

carrotIndustries:
> I like that idea! I guess these could be inherited from other pools or overwritten, but how would pools inheriting from multiple other pools choose them?

Yes, these will be able to be overridden. For multiple included pools, the first one will win.

carrotIndustries:
0117ddb95d9614409ff1a3d1a1c425044b9cbbb6 adds loading layer help texts from the pool
To make overriding behaviour consistent with pool items, the last included pool will win.

Closing this, please open a new PR in the pool.

Issue #206, opened 2018-11-18
[UI] Allow fine grid movement with Alt + Arrow


atoav:
In #154 I proposed the general arrow key movement of parts, which I happen to use extensively since then (probably 50% of all movements I do). A minor detail that I also proposed back then was to allow the use of the same Alt-Modifier you implemented for the mouse move command <kbd>m</kbd>.  

The idea was to use <kbd>Alt</kbd> + <kbd>←/→/↑/↓</kbd> to move things on the finer grid via keyboard.  

This is obviously not very high priority, but I'd still love to see this, because keyboard movement especially shines when you have to move stuff just a tiny bit.

carrotIndustries:
Would starting the move tool via Alt-arrow resulting in all subsequent arrow keys being interpreted as fine movements work for you?

atoav:
To be honest I'd like it more if this would be global in a similar fashion like <kbd>←</kbd>, <kbd>↑</kbd>, <kbd>→</kbd> and <kbd>↓</kbd> are at the moment. That means while <kbd>Alt</kbd> is beeing pressed every arrow key would move on a fine grid (just like it moves things on the coarse grid without <kbd>Alt</kbd>). This way you can alternate quicker between fine and coarse grid movement, which is IMO more intuitive.

But I suspect this would be harder to implement than your proposed solution? If this is a problem, then just go ahead and implement it the way you described.

atoav:
So I am not the only one : )

@frmdstryr You know you can use math operations in the grid field too? e.g. if you type `/10` after the existing value he should divide by ten and if you type `*10` he should multiply by ten

carrotIndustries:
The reason why this has been stalled for so long is that I haven't yet figured out a satisfactory way to implement this. Tools/actions are activated by the key sequences defined by the user, adding move fine left/right/up/down would require the user to change the key sequences in sync with the fine grid modifier to maintain consistency. We'll also need to figure out whether to perform a fine move operation in the tool itself. Right now, a tool doesn't really know about preferences.

atoav:
I am not sure about this, but what if instead of changing the tool, <kbd>Alt</kbd> would change the environment, unless used in a combination specified in the users preferences. 

That means pressing <kbd>Alt</kbd> could switch the grid from coarse to fine, while the arrow keys just Move by one grid unit, whatever it may be at the moment.

[...]

PR #230, opened 2019-02-04
UI: Add coarse and fine keyboard movements

frmdstryr:
Adds the `mod` to ToolArgs key presses which can be used to distinguish between the different move modes (fine, default, coarse).  For #206 

The only todo is that the modifier binding isn't configurable, other than that it works quite nicely.


carrotIndustries:
> perhaps it'd be better to generate actions based on the preferences?

I've though a bit about more about that and came up with that plan:

Having the key sequences for the "move keyboard" tools configurable is a bit silly. No one's going to change these and those who do will be wondering why things broke. To remove these from the preferences dialog, add a new flag `NO_PREFERENCES` to `ActionCatalogItem::Flags`, then set the key sequences for these tools depending on the users' preference for the fine modifier.

> I was thinking of re-checking for any matching key sequences to lookup the tool id, in
[`bool ImpBase::handle_key_press(GdkEventKey *key_event)`](https://github.com/carrotIndustries/horizon/blob/master/src/imp/imp.cpp#L1040) then assign the matching tool_id to the ToolArgs.

Since I can't yet think of a use case apart from the move tool, I'm not really convinced by this.

To know whether to do a fine movement in `ToolMove::update` figure out the modifiers in the imp and pass normal/fine to the tool.

frmdstryr:
What about just doing this: https://github.com/carrotIndustries/horizon/compare/master...frmdstryr:move-tool

It works good enough, the only somewhat confusing thing is that the move tool must be activated first before a fine move can be done. It's reusing the `button` but it's probably better to add a new member to the ToolArgs. 

Or it may just be better for you to implement it however you think is best :+1: (feel free to just close this)

carrotIndustries:
> It works good enough, the only somewhat confusing thing is that the move tool must be activated first before a fine move can be done.

That's not really what we want.

> Or it may just be better for you to implement it however you think is best +1 (feel free to just close this)

done in 3b12b58e34cb1b9fcbe97f4e6e624a4c895b09d6

Thanks for making me think a bit more about this issue!

frmdstryr:
Works great! Thank you!

PR #236, opened 2019-02-25
Implement Q3DO Import

q3k:
This PR implements support for the [Q3D(o)](https://github.com/q3k/q3d/) 3d mesh format.

We ship FlatBuffers and Q3D as git submodules, add Makefile machinery, and implement the import functionality.

A sample file for testing: https://q3k.org/u/2f0f39a94b6e7bd47e5efb689b6c84e3d55e3435fd6c3627cf834da4190e5dc9.q3do (ESP32-WROOM-32 package)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,9 @@ jobs:
           name: Install dependencies
           command: apt-get update -y && apt-get install libyaml-cpp-dev libsqlite3-dev util-linux librsvg2-dev libcairomm-1.0-dev libepoxy-dev libgtkmm-3.0-dev uuid-dev libboost-dev  libzmq5 libzmq3-dev libglm-dev libgit2-dev libcurl4-gnutls-dev liboce-ocaf-dev libpodofo-dev python3-dev libzip-dev git -y
       - run:
+          name: Check out submodules
+          command: git submodule init && git submodule update
+      - run:
           name: Build
           command: make -j2
       - run:
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ horizon-pool-mgr
 horizon-rest-test
 horizon-gen-pkg
 horizon-eda
+flatc
 rest-test.cpp
 *.o
 *.oshared
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "3rd_party/q3d"]
+	path = 3rd_party/q3d
+	url = https://github.com/q3k/q3d
+[submodule "3rd_party/flatbuffers"]
+	path = 3rd_party/flatbuffers
+	url = https://github.com/google/flatbuffers
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PKGCONFIG=pkg-config
 BUILDDIR = build
 PROGS    = $(addprefix $(BUILDDIR)/horizon-,imp pool prj eda)
+FLATC=$(shell pwd)/flatc
 
 all: $(PROGS)
 pymodule: $(BUILDDIR)/horizon.so
@@ -94,7 +95,7 @@ SRC_COMMON = \
 	src/frame/frame.cpp\
 	src/common/keepout.cpp\
 	src/board/board_layers.cpp\
-	src/pool/pool_parametric.cpp \
+	src/pool/pool_parametric.cpp\
 
 ifeq ($(OS),Windows_NT)
     SRC_COMMON += src/util/uuid_win32.cpp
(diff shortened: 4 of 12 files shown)


q3k:
Sister PR in SolveSpace: https://github.com/solvespace/solvespace/pull/384

atoav:
Interesting! Is there any good way to go from q3Do to Blender?

q3k:
Not yet, but writing a blender importer should be fairly easy for someone who's familiar with their API. FlatBuffers (which are the main 'transport' layer of q3do) are supported in Python.

atoav:
There file format is not that straight forward, but the API is very usable. The best solution would probably be to write a plugin. Never did that tho, just some scripting.

Visualization is a good thing. The step format is good enough for the internal 3D-viewer, but exporting parts with color and all that jazz to visualize it with a a physically more accurate renderer like Blender's Cycles is incredibly hard because you have to convert step to different other formats as Blender does not support it. And then you end up having no colors manually recoloring 40 SMD resistors is a length task that you would have to redo as soon as something changes. 

That brings me to the next thing: I used solidpython a ton for this project. solidpython is a python wrapper around OpenSCAD that helps to quickly generate tons of variants of parts. I am not familiar with Q3DO, but the stuff I saw could lend itself quite easily to similar approaches.

carrotIndustries:
TBH, I'm still not quite sure whether we actually want mesh import in horizon:

Con:
- meshes are only useful for display purposes, can't be used in STEP export. Having a well-working export is more important (to me at least) than a pretty 3D preview, horizon is a CAD application after all.
- supporting both STEP and mesh files for one model adds significant complexity (in terms of UI and code)

Pro:
 - mesh files are easier to generate correctly than STEP

carrotIndustries:
Sorry for letting this sit for so long, my primary issue with this PR isn't so much the actual code added to horzion itself, but rather the added complexity in terms of submodules and additional build steps needed for a rather niche feature.

If we already had submodules with all the required tooling to build proper source tarballs in place, I'd probably already have merged this PR as it'd just add some more submodules to the list.

q3k:
No worries, I'm not in a rush or married to this PR either - I just rebased it on top of master and pushed for others to use if necessary.

May I interest you in a better build system like Bazel? :)

carrotIndustries:
Build systems can be a surprisingly touchy topic, see #111  From my current perspective, there aren't that many options if one doesn't want to displease maintainers too much:

 - Make (as is)
 - cmake (if it ends up less complex than the proposal in #111, but cmake's syntax is still too obtuse to my liking)
 - autotools (ugh, but at least people know it)
 - meson

bazel is a non-starter as it requires java which seems a bit heavy just to build software.

From all of the options above, meson would be my preferred option if there'd be good reasons to use something else than plain makefiles as it seemed to have replaced autotools in the gtk universe.

q3k:
> bazel is a non-starter as it requires java which seems a bit heavy just to build software.

Not an entire JRE at least - it comes with its own, hermetic, minified JRE within the installer, which weighs a total of:
```
q3k@anathema ~ $ du -hs bazel-1.2.1-installer-linux-x86_64.sh 
43M	bazel-1.2.1-installer-linux-x86_64.sh
```

FWIW, I'm not a huge fan of Java either, but I've never felt that Bazel is heavy because of it.

q3k:
You also have to remember Bazel gives you a whole bunch of features over a plain make or most other build systems:

 - hermetic/reproducible builds
 - machine-wide and distributed caching of builds _and_ tests (fantastic for CI)
 - distributed builds
 - import of external dependencies via HTTP of VCSes (poignant to the comment that started this)
 - a pretty well developed, reusable ecosystem for languages and other tooling

I can volunteer to make an experimental bazelification branch with no strings attached :).

carrotIndustries:
For build systems, popularity is also a somewhat important measure when it comes to not displeasing package maintainers.

On arch linux at least bazel just seems to be in the repos because of tensorflow https://www.archlinux.org/packages/community/x86_64/bazel/ whereas meson is used to build close to 400 packages. Debian for example doesn't package bazel at all. This makes bazel a non-viable build system for this project.

q3k:
I can't say I agree with this argument, but it's your project.

I assume this also makes this PR irrelevant for now - should I close it?

carrotIndustries:
> I assume this also makes this PR irrelevant for now - should I close it?

As I don't see any perspective of getting this PR merged in the near future, we're probably better of closing it for now.

If the build system changes in a favorable way in the future, we can rethink merging this PR.

ruevs:
In the meantime SolveSpace now exports Step files with color that Horison displays correctly :-)

https://github.com/solvespace/solvespace/pull/767

https://github.com/solvespace/solvespace/pull/763

https://github.com/solvespace/solvespace/issues/452

PR #260, opened 2019-05-11
Show polygon edge length when a single edge is selected

frmdstryr:
horizon-polygon-edge-length


carrotIndustries:
Is there any reason for limiting this to a single edge?

frmdstryr:
I just wanted to verify some dimensions that were not lines. Showing the length of more than one only made sense to me if the selected edges are all continuous segments of the same polygon and I wasn't sure if I wanted to check for that (or how to display it properly) so I kept it simple.

carrotIndustries:
Ok, in that case, I suggest changing "Total length" to "Length".

PR #262, opened 2019-05-14
Fix icon on wayland

frmdstryr:
On wayland the program name needs to be set or it splits the schematic into separate taskbar icons.

Before:
Screenshot_20190514_103638

After:
Screenshot_20190514_110001

### DIFF
--- a/src/imp/imp.cpp
+++ b/src/imp/imp.cpp
@@ -667,6 +667,8 @@ void ImpBase::run(int argc, char *argv[])
 
     Gtk::Window::set_default_icon_name("horizon-eda");
 
+    Glib::set_prgname("horizon-eda"); // Fixes icons on wayland
+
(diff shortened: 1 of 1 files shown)


carrotIndustries:
I don't use non-tiling window managers, but for me having the schematic/board editor as a separate button in the task bar seems like a plus to me as it makes switching to that window easier. The icon definitely needs fixing though.

frmdstryr:
The stacking must be a wayland only thing (or a config difference) because if I switch to an X session it's still split into two taskbar items with this change.

carrotIndustries:
Ok, since I don't have a non-tiling window manager on wayland to test things, I'll merge it as the change isn't that intrusive.

If someone complains they don't like the way it behaves on their desktop environment, we can revisit this issue.

Issue #269, opened 2019-05-20
Feature request: Zip fabrication outputs


frmdstryr:
Would be nice to have a checkbox to write the fabrication outputs to a zip file.

frmdstryr:
Added in https://github.com/horizon-eda/horizon/pull/270

PR #270, opened 2019-05-22
Write output files to zip for #269

frmdstryr:


carrotIndustries:
LGTM overall, some minor UI nitpicks
 - The zip output setting isn't stored and loaded from the settings (use `bind_widget`)
 - The zip checkbox looks a bit odd at it's current place. A switch integrates better with the rest of the window:
2019-05-22-221005_372x158_scrot

src/imp/fab_output_window.cpp
```diff
@@ -75,6 +76,7 @@ FabOutputWindow::FabOutputWindow(BaseObjectType *cobject, const Glib::RefPtr<Gtk
     prefix_entry->signal_changed().connect([this] { s_signal_changed.emit(); });
     bind_widget(npth_filename_entry, settings->drill_npth_filename, [this](std::string &) { s_signal_changed.emit(); });
     bind_widget(pth_filename_entry, settings->drill_pth_filename, [this](std::string &) { s_signal_changed.emit(); });
+    bind_widget(zip_output_switch, settings->zip_output);
```
carrotIndustries: add `[this](std::string &) { s_signal_changed.emit(); }` as for the widgets above to update the board's save status

Issue #271, opened 2019-05-24
macOS Support


phlmn:
I think it's a good idea to track the state of macOS support inside an issue.

I started to implement macOS support with #264, but are is quite some things missing. Things I currently have in my mind:

- [x] make it compile (#264)
- [x] make it start (#264)
- [ ] make the board viewer work
- [ ] fix icons

phlmn:
## OpenGL Issue (from #264)

Unfortunately the board viewer does not work yet. glLineWidth(…) and glBlitFrameBuffer(…) are causing an error. By removing glLineWidth(…) and drawing directly to the screen instead of using glBlitFrameBuffer(…) the board viewer does start. Debugging OpenGL is not much fun, maybe someone has an idea.

Some potentially useful apitrace output:

```
$ ./apitrace replay horizon-imp.5.trace
/Users/phlmn/Coding/apitrace/build/../../horizon/horizon-imp
1 @0 CGLRetainPixelFormat(pix = 0x7ff682f98870) = 0x7ff682f98870
1: warning: unsupported CGLRetainPixelFormat call
2 @0 CGLReleasePixelFormat(pix = 0x7ff682f98870)
2: warning: unsupported CGLReleasePixelFormat call
4 @0 CGLRetainContext(ctx = 0x7ff68406a400) = 0x7ff68406a400
4: warning: unsupported CGLRetainContext call
5 @0 CGLReleaseContext(ctx = 0x7ff68406a400)
5: warning: unsupported CGLReleaseContext call
6 @0 CGLReleasePixelFormat(pix = 0x7ff682f98870)
6: warning: unsupported CGLReleasePixelFormat call
14 @0 CGLRetainPixelFormat(pix = 0x7ff682e7a510) = 0x7ff682e7a510
14: warning: unsupported CGLRetainPixelFormat call
15 @0 CGLReleasePixelFormat(pix = 0x7ff682e7a510)
15: warning: unsupported CGLReleasePixelFormat call
17 @0 CGLRetainContext(ctx = 0x7ff68492f800) = 0x7ff68492f800
17: warning: unsupported CGLRetainContext call
18 @0 CGLReleaseContext(ctx = 0x7ff68492f800)
18: warning: unsupported CGLReleaseContext call
19 @0 CGLReleasePixelFormat(pix = 0x7ff682e7a510)
19: warning: unsupported CGLReleasePixelFormat call
GL_RENDERER: Intel(R) Iris(TM) Graphics 550
[...]

PR #264, opened 2019-05-14
macOS Support (frist steps)

phlmn:
Hey,
I managed to get something working on macOS.

For this I had to:
- update nlohmann json
- update the Makefile
- import libuuid from `uuid/uuid.h` instead of `uuid.h`
- write `get_exe_dir()` for macOS

Unfortunately the board viewer does not work yet. `glLineWidth(…)` and `glBlitFrameBuffer(…)` are causing an error. By removing `glLineWidth(…)` and drawing directly to the screen instead of using `glBlitFrameBuffer(…)` the board viewer does start. Debugging OpenGL is not much fun, maybe someone has an idea.

Some potentially useful apitrace output:
```
$ ./apitrace replay horizon-imp.5.trace
/Users/phlmn/Coding/apitrace/build/../../horizon/horizon-imp
1 @0 CGLRetainPixelFormat(pix = 0x7ff682f98870) = 0x7ff682f98870
1: warning: unsupported CGLRetainPixelFormat call
2 @0 CGLReleasePixelFormat(pix = 0x7ff682f98870)
2: warning: unsupported CGLReleasePixelFormat call
4 @0 CGLRetainContext(ctx = 0x7ff68406a400) = 0x7ff68406a400
4: warning: unsupported CGLRetainContext call
5 @0 CGLReleaseContext(ctx = 0x7ff68406a400)
5: warning: unsupported CGLReleaseContext call
6 @0 CGLReleasePixelFormat(pix = 0x7ff682f98870)
6: warning: unsupported CGLReleasePixelFormat call
14 @0 CGLRetainPixelFormat(pix = 0x7ff682e7a510) = 0x7ff682e7a510
14: warning: unsupported CGLRetainPixelFormat call
15 @0 CGLReleasePixelFormat(pix = 0x7ff682e7a510)
15: warning: unsupported CGLReleasePixelFormat call
17 @0 CGLRetainContext(ctx = 0x7ff68492f800) = 0x7ff68492f800
17: warning: unsupported CGLRetainContext call
18 @0 CGLReleaseContext(ctx = 0x7ff68492f800)
18: warning: unsupported CGLReleaseContext call
19 @0 CGLReleasePixelFormat(pix = 0x7ff682e7a510)
19: warning: unsupported CGLReleasePixelFormat call
GL_RENDERER: Intel(R) Iris(TM) Graphics 550
GL_VENDOR: Intel Inc.
477 @0 glLinkProgram(program = 8)
[...]

PR #272, opened 2019-05-25
Update JSON Library

phlmn:
In order to make this compile in macOS I updated nlohmann JSON to the newest version.

Splitted out of PR \#264.

Part of #271.


carrotIndustries:
To which version did you update the json library?
Also seems like we were relying on some includes that we now need to include ourselves.

carrotIndustries:
done in 87566ed5c6ef804de2c2f694017a8b3b28141402

PR #275, opened 2019-05-25
Fix random things for macOS

phlmn:
Splitted out of \#264.

Part of \#271.


carrotIndustries:
fixed in deee6a58c8f82b39befe175110177abc6dc7294c

PR #279, opened 2019-05-30
Write build output to separate dir

endofexclusive:
# Summary
This pull request allows writing compiler output files to an out-of-source directory. It can be useful when comparing behavior of the program in different branches for example. Also keeps source tree listings a bit cleaner.

The intention is to not touch the source directory when building. A separate build tree is created and populated with object files. It has the same directory structure as the source tree.

# Examples
Builds in a directory named "build":
`make`
    
Build in "build-nodebug" without debugging information:
` make DEBUG= BUILDDIR=build-nodebug`
    
Build in a directory named by the git tag:
``` make BUILDDIR=build-`git describe --always --tags` ```
    
Build in source tree (old behavior):
`make BUILDDIR=. OBJDIR=.`

This pull request also decreases build verbosity by default. `make QUIET=` can be used to get verbose output with full compiler command lines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
           command: make -j2
       - run:
           name: Build python module
-          command: make -j2 horizon.so
+          command: make -j2 build/horizon.so
 
 workflows:
   version: 2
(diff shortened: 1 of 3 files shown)


carrotIndustries:
While we're at it, we might as well fix the ugly hack that's used to build the python module at the moment:

Since the source files for the module have to be built with `-fPIC`, they're built separately (`.oshared`) unfortunately, the header dependency files still have the same name which might cause conflicts if both happen to be built at the same time.

By building the python module in a separate directory, this issue can easily be solved.

endofexclusive:
Now the python module objects are built in a separate directory and are being removed by the `clean` target. The second commit reduces some repetition by using target specific include directories.

PR #287, opened 2019-07-01
Footprint generator with IPC-7351B etc.

endofexclusive:
This adds an interactive footprint generator based on IPC-7351B.
screen0

It consists of two main parts
- A library for generating footprints (footag)
- Integration in Horizon

# The library
The library contains a catalog of component types and knowledge on how to generate land patterns for these components. As part of the application interface, the library describes what kind of parameters a component type has.

It is up to the application do display and fill in the parameters. When the application has filled in the parameters, the library is ready to calculate a land pattern based on the parameters.

The library has no dependencies and no graphical representation.

# Integration in Horizon
The "Footprint gen." has got a new panel which presents the different component types known by the library, together with its announced parameters. The program has no knowledge of the calculations and component features: what parameters to show in the GUI is provided to Horizon at run-time and rendered into appropriate Gtk widgets.

A preview canvas shows what the package will look like.

The values next to the ± button represents nominal value and tolerance as plus/minus value. By clicking the ± button, the widget instead becomes a min/max range.

# Calculations
The calculations represent my understanding of IPC-7351B.

There are also some component types (through-hole) which are based on "IPC-7251B 1st working draft". That is basically a combination of the through-hole recommendations in IPC-2221 and IPC-2222.

# About IPC-7351B calculations
A land pattern is calculated based on the following parameters:
- Component lead dimensions
- Component lead tolerances
- PCB fabrication tolerance
- Part placement tolerance
- Density level

All of these can be controlled in this wizard.

# Limitations
There is no feedback path from user input values to what widgets to display. Each widget is constructed only once.

The Horizon integration needs some work. For example to use `well_known_padstack` semantics.

# Others
- IPC-7351B does not specify what artwork shall look like. So things like silkscreen is not covered.
- It would be interesting to add a few parameters (height) to each component type to be able to generate 3D models.
- Some kind of graphical legend which describes the different parameters would be useful. The best guide at the moment is to click the gadgets and see what happens in the preview.

# Links
An excellent introduction to IPC-7351B, published by Mentor Graphics:
https://www.innofour.com/3783/news/literature/pcb-design-perfection-starts-in-the-cad-library/pcb-design-perfection-starts-in-the-cad-library-part-1

Some thoughts about the current state of IPC-7351 by one of the committee members:
https://www.pcblibraries.com/forum/ipc7351c-draft-or-release-date_topic1818_page2.html


carrotIndustries:
Awesome, integrating the well known padstacks should be a piece of cake.

I guess you won't mind if I have a go at bringing the widget layout in line with the rest of the application and the GNOME guidelines.

endofexclusive:
No problem!

carrotIndustries:
See https://github.com/horizon-eda/horizon/tree/ipc-wizard for my changes.

2019-07-04-005636_1197x796_scrot

carrotIndustries:
Do you plan on adding the `well_known_padstack` integration or shall I have a go at it? Do you suspect any issues with it or are there padstacks that still need a well known name?

Is there anything more to fix before merging this?

endofexclusive:
The updated GUI looks very good!

I have nothing special in mind for the `well_known_padstack`. The only thing to consider would be if the padstack can not be found, however I believe the current behavior is reasonable (log a warning that describes the situation). Please do the `well_known_padstack` if it is OK for you.

One thing I was thinking of doing is to make the attribute descriptions and "hints" more complete. But that can be done after a merge.

There are currently no changes for the footprint API in the pipe from my side. So if you are happy with the wizard, then let's merge.

carrotIndustries:
Merged in c6f42667f3ba995bd29809c3cb7893781a172570, d04a3fc9c22dd67ee41de8dc4b18fcc8a2a5db6e, 0c07d8e729ff1483c26f365d0d90ce68edc2a332

PR #290, opened 2019-08-17
build and run on osx

andryblack:
Fix building and running issues for MacOSX. 
brew dependencies: gtkmm3,opencascade,libzip,yaml-cpp,osrf/homebrew-simulation/cppzmq, podofo,boost,librsvg,glm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -573,9 +573,10 @@ ifneq ($(OS),Windows_NT)
 endif
 LIBS_ALL = $(LIBS_COMMON) gtkmm-3.0 epoxy cairomm-pdf-1.0 librsvg-2.0 libzmq libgit2 libcurl glm
 
-OPTIMIZE=-fdata-sections -ffunction-sections
+OPTIMIZE=-fdata-sections -ffunction-sections -O3
+#OPTIMIZE=-fdata-sections -ffunction-sections
 DEBUG   =-g3
-CXXFLAGS  =$(DEBUG) $(DEFINES) $(OPTIMIZE) $(shell $(PKGCONFIG) --cflags $(LIBS_ALL)) -MP -MMD -pthread -Wall -Wshadow -std=c++14 -O3
+CXXFLAGS  =$(DEBUG) $(DEFINES) $(OPTIMIZE) $(shell $(PKGCONFIG) --cflags $(LIBS_ALL)) -MP -MMD -pthread -Wall -Wshadow -std=c++14 
 CFLAGS = $(filter-out -std=%,$(CXXFLAGS)) -std=c99
 LDFLAGS = -lm -lpthread
 GLIB_COMPILE_RESOURCES = $(shell $(PKGCONFIG) --variable=glib_compile_resources gio-2.0)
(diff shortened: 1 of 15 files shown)


carrotIndustries:
Cool, to make sure that it stays that way, we'd need some sort of CI for Mac OS. Are there any free CI offerings for Mac OS?

andryblack:
[Travis](https://docs.travis-ci.com/user/reference/osx/) has Mac OS environment. I can try add configuration.

andryblack:
Added Travis CI configuration and script for building movable Mac application.

carrotIndustries:
It'd be great if you could split this up into several commits before merging this so that we can easily find out if it broke things:

 - json fixes (#275 mentioned that updating the bundled json.hpp might do it)
 - opengl related
 - bundling
 - CI

andryblack:
I can do it on the weekend

carrotIndustries:
Is Horizon EDA actually usable on OS X with these patches apart from "it compiles and runs". Given the deprecation (?) of OpenGL on mac os and the required app notarisation starting with catalina, I'm to entirely sure if supporting mac os is worth it on the long run. Any comments?

bkurtz:
@carrotIndustries You might consider a hybrid approach where you keep the code in a format that's more-or-less compatible with macOS (e.g. the change that seems to have made it in to allow the executable to detect its location in the filesystem), and allow actual packaging/distribution to be handled by, e.g. [MacPorts](https://macports.org) or [Homebrew](https://brew.sh).

WRT OpenGL support in macOS, it seems like at least for the time being, they're keeping it around but just not updating it, so as long as you don't need the newest shiniest features, things should mostly be okay (I think).

I currently have a recipe for Horizon on Homebrew, which builds it successfully and it runs and does some of the basic UI stuff, but board and schematic editing are broken due to an OpenGL issue.  I'm still troubleshooting the OpenGL issue, but I think it actually just boils down to that GTK does a check for a particular feature and Apple decided when doing their OpenGL implementation that as that feature is part of the GL core that they don't include it in their list of supported features, because "it's part of core now, so obviously we support it" (nevermind that Khronos Group recommends including it in the feature list anyway).  I'm still working on verifying that it works if the check is disabled, but Apple's docs suggest the feature *should* be supported, so I'm optimistic.

The other good news would be that my current Homebrew recipe is based off of your `master` branch (or rather, off of your `1.0.0` tagged release), so there shouldn't be any additional changes required.

bkurtz:
Well, on second thought, it looks like it's going to be quite a lot of work to get this going using the version of Apple OpenGL (3.3) on my Macbook, so I'm going to give up on this for today.

arjenroodselaar:
Hi @bkurtz, have you been able to make any progress on this? I would love to help figure out if we can make this work.

bkurtz:
@arjenroodselaar I was working on a build script for homebrew.  Have a look at bkurtz/homebrew-core@0210adc2e64506e39233b941fc22f8fef96ae45a.  I can't recall for sure, but *I think* I got it to the point that it would build and launch, but then most of the buttons in the UI wouldn't do anything because my macbook is quite old (2011!) and apple hasn't updated the version of OpenGL on it recently enough I guess.

PR #318, opened 2019-11-14
make 'make clean' really clean

### LINKED ISSUE #316: make clean is not clean enough
There are still some leftovers after 'make clean'. I suggest to add $OBJ_COMMON to the 'rm' in the clean target. But even then, there is still a __pycache__ dir and some *.d files in build/obj/.... I actually don't know how to get rid of.

carrotIndustries:
fixes #316

@endofexclusive  does that look about right to you?


endofexclusive:
Yes it looks good.

PR #325, opened 2019-12-05
Angle spinbox improvements

fruchti:
I noticed the angle spinbox didn't take decimal places when the locale's decimal separator is a comma. While fixing this I changed a few other things:

* The increments with the arrow keys/scrolling are now 1°, with PgUp/PgDn 15°. This should be more useful than the current behaviour.
* You can now directly enter negative angles and angles > 360°. Especially negative angles are useful sometimes.

Because I did neither find a way to let Gtk consider the upper and lower limit to be identical, nor how to easily check if the last increment was a step or a page increment, the wrapping behaviour is now slightly inconsequential: with both kinds of increments, it always wraps to 0° or 359°. I consider this issue to be pretty minor, though.

### DIFF
--- a/src/property_panels/property_editor.cpp
+++ b/src/property_panels/property_editor.cpp
@@ -6,6 +6,7 @@
 #include "core/core_schematic.hpp"
 #include "core/core_board.hpp"
 #include "widgets/spin_button_dim.hpp"
+#include "widgets/spin_button_angle.hpp"
 #include "util/util.hpp"
 #include <algorithm>
 #include <iostream>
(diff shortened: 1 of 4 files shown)


fruchti:
Regarding the failed build:

```
 build/gen/version_gen.cpp
Traceback (most recent call last):
  File "make_version.py", line 25, in <module>
    fi.write(tmpl.substitute(major=v.major, minor=v.minor, micro=v.micro, string=v.string, name=v.name, commit=gitversion))
UnicodeEncodeError: 'ascii' codec can't encode character '\xb0' in position 304: ordinal not in range(128)
Makefile:704: recipe for target 'build/gen/version_gen.cpp' failed
make: *** [build/gen/version_gen.cpp] Error 1
make: *** Waiting for unfinished jobs....
```

Did the ‘°’ in the commit message break make_version.py? It builds just fine on my system…

carrotIndustries:
Pressing page down if the angle is 0 results in 358.99° rather than 345

fruchti:
Yes, I would have preferred wrapping to 360° - step size, too, but I didn't find a nice way to achieve this. I'll look into somehow reimplementing the wrapping manually if it's a big issue for you.

fruchti:
Found a way around it. Wrapping should now behave as intended.

PR #335, opened 2020-01-16
SpinButtonDim parses mil

endofexclusive:
Allows typing things like `4.5 mil` (in addition to `mm` and `inch`) into dimension spin buttons. It is not the neatest patch, but it works.


carrotIndustries:
While working on that code, I got the feeling that this piece of code might be good reason to add some tests...

endofexclusive:
Thanks!

PR #345, opened 2020-02-04
Add component group and tag to schematic and board HUD

fruchti:
This PR adds the group and tag to the schematic editor's HUD:
screenshot

This is especially useful if a symbol's value is hidden or the schematic if is too crowded, so the group and tag names displayed in the value field wouldn't be readable.

Am I right in assuming that adding this information to the board editor would be non-trivial, since the block and consequently the group and tag names aren't available there?

### DIFF
--- a/src/imp/imp.cpp
+++ b/src/imp/imp.cpp
@@ -924,6 +924,12 @@ std::string ImpBase::get_hud_text_for_component(const Component *comp)
             s += "<a href=\"" + Glib::Markup::escape_text(part->get_datasheet()) + "\" title=\""
                  + Glib::Markup::escape_text(Glib::Markup::escape_text(part->get_datasheet())) + "\">Datasheet</a>\n";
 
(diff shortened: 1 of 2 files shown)


carrotIndustries:
> Am I right in assuming that adding this information to the board editor would be non-trivial, since the block and consequently the group and tag names aren't available there?

In fact, simply move you your code over to ` ImpBase::get_hud_text_for_component` and it'll automatically work in both schematic and board. Use `core.r->get_block()` to get the netlist. Also use `.at()`. rather than `[]` and make the pointer to the block `const` to avoid accidentally modifying it.

fruchti:
Oh, neat! In fact, I had the code in `ImpBase` at first, but used `core.c` to try to access the block which understandably fails for the board editor.

carrotIndustries:
One more thing: use `Block::get_group_name` and its `tag` friend to get the string representations.

fruchti:
Done.

PR #347, opened 2020-02-19
add "do not populate" component marking

eqvinox:
Only in the schematic editor for now;  marks an entire component (with
all its symbols) as DNP.  Drawn grayed out with an X over it (calculated
from the bounding box, occasionally looks a bit weird but eh.)

TBD/Future:
- ~add BOM export option to exclude DNP parts~ done
- visualize in board editor & assembly export

screenshot for posterity:
https://aurora.nox.tf/tmp/horizon_dnp.png

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,6 +236,7 @@ SRC_IMP = \
 	src/core/tools/tool_smash_silkscreen_graphics.cpp\
 	src/core/tools/tool_renumber_pads.cpp\
 	src/core/tools/tool_fix.cpp\
+	src/core/tools/tool_nopopulate.cpp\
 	src/core/tools/tool_polygon_to_line_loop.cpp\
 	src/document/documents.cpp\
 	src/core/clipboard.cpp\
--- a/src/block/block.cpp
+++ b/src/block/block.cpp
@@ -303,6 +303,9 @@ std::map<const Part *, BOMRow> Block::get_BOM(const BOMExportSettings &settings)
 {
     std::map<const Part *, BOMRow> rows;
     for (const auto &it : components) {
+        if (it.second.nopopulate && !settings.include_nopopulate) {
+            continue;
+        }
         if (it.second.part) {
             const Part *part;
             if (settings.concrete_parts.count(it.second.part->uuid))
--- a/src/block/bom_export_settings.cpp
+++ b/src/block/bom_export_settings.cpp
@@ -11,7 +11,8 @@ const LutEnumStr<BOMExportSettings::CSVSettings::Order> bom_order_lut = {
 };
 
 BOMExportSettings::BOMExportSettings(const json &j, Pool &pool)
-    : csv_settings(j.at("csv_settings")), output_filename(j.at("output_filename").get<std::string>())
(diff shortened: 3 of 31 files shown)


carrotIndustries:
Cool! It's really reassuring to know that other people than me can make non-trivial changes as well.

Some minor things:

I think that drawing symbols in grey and adding the cross is overdoing it bit. The cross by itself is probably fine. I've had in mind to use the grey style to indicate components that aren't yet placed on the board at some point in the future.

A simple tool to set/clear the unpopulate flag would be nice as it allows to modify this flag more directly (context menu or shortcut). See https://github.com/horizon-eda/horizon/blob/master/src/core/tools/tool_fix.cpp for a similar tool.

eqvinox:
updated; I left the code for drawing symbols in a different colors in a separate commit so you can reuse it for the "not placed yet" indication.

> Cool! It's really reassuring to know that other people than me can make non-trivial changes as well.

That's pretty funny because I'm not actually a C++ developer ;D.  Then again I would consider this PR a trivial change really.

eqvinox:
added BOM export checkbox for DNP parts

eqvinox:
Not sure why the AppVeyor CI says "Build has been cancelled by a user.", at least I didn't do anything there.

carrotIndustries:
LGTM, as far as I can tell, the only real thing missing is supporting the new flag in the pick&place export?

carrotIndustries:
One more thing: A common feature among more upmarket EDA software is support for variants.

Suppose you're building something that could be made as either a one or two channel variant. By supporting variants we could represent this in the design data.

Unfortunately, this gets rather unwieldy if we have two or more dimensions that could be varied independently, such as channel count and number of memory chips populated. To elegantly solve this case, we'd need something like meta-variants: One for channel count, one for the memory. The meta variants could then be merged into actual variants that'll get made.

Anyhow, the simple 'do not place' flag is probably enough for most people's use cases.

eqvinox:
> LGTM, as far as I can tell, the only real thing missing is supporting the new flag in the pick&place export?

Oh yeah I totally forgot about that since I haven't used it yet.  I can probably wire that up quickly.

What's also left to do is board/assembly view support; I have a hack but I'm not particularly happy about it and I think this is something for the "next" PR.

> One more thing: A common feature among more upmarket EDA software is support for variants.

I think a good way of doing this without getting overly complex might be with plain old tagging things.  Independent tags can represent orthogonal "dimensions" easily, and if needed there can be an analog of "design rules" that describe tag interdependencies (e.g. "channel3" tag requires "channel2" tag, "230V" tag conflicts with "115V" tag, etc.)

eqvinox:
(add:) Good "Variant" support probably needs some sensible way of swapping out different parts in the same footprint, e.g. for different resistor values or even different IC variants...

eqvinox:
PNP export now has a checkbox like BOM export.

Also, thanks for the quick PDF export fix :)

carrotIndustries:
I think what we have at this point is a good enough for most use cases, so I merged it.

PR #352, opened 2020-03-18
Use two finger gesture on touchpads for panning instead of zooming

### LINKED ISSUE #350: Add option for panning instead of zooming with the touchpad
Hey,
first of all thanks for the great work on horizon eda in general and on the nice input especially.
It would be great for touchpad usability if there was an option to use two fingers on touchpads for panning instead of zooming. 

Panning with the touchpad already works, when did two-finger zoom ("pinch") gesture was recognized and is active. Having this mode as the default behaviour would be great.

anuejn:
closes #350

### DIFF
--- a/src/canvas/canvas_gl.cpp
+++ b/src/canvas/canvas_gl.cpp
@@ -273,7 +273,7 @@ bool CanvasGL::on_scroll_event(GdkEventScroll *scroll_event)
 #if GTK_CHECK_VERSION(3, 22, 0)
     auto *dev = gdk_event_get_source_device((GdkEvent *)scroll_event);
     auto src = gdk_device_get_source(dev);
-    if (src == GDK_SOURCE_TRACKPOINT) {
+    if (src == GDK_SOURCE_TRACKPOINT || (src == GDK_SOURCE_TOUCHPAD && touchpad_pan)) {
         if (scroll_event->state & GDK_CONTROL_MASK) {
             pan_zoom(scroll_event, false);
         }
--- a/src/preferences/preferences_util.cpp
+++ b/src/preferences/preferences_util.cpp
@@ -14,10 +14,12 @@ void preferences_provider_attach_canvas(CanvasGL *ca, bool layer)
         a = &PreferencesProvider::get_prefs().canvas_non_layer.appearance;
     }
     ca->set_appearance(*a);
+    ca->touchpad_pan = PreferencesProvider::get_prefs().zoom.touchpad_pan;
     PreferencesProvider::get().signal_changed().connect(sigc::track_obj(
             [ca, a] {
                 ca->set_appearance(*a);
                 ca->smooth_zoom = PreferencesProvider::get_prefs().zoom.smooth_zoom_2d;
+                ca->touchpad_pan = PreferencesProvider::get_prefs().zoom.touchpad_pan;
             },
             *ca));
 }
--- a/src/canvas/canvas_gl.hpp
+++ b/src/canvas/canvas_gl.hpp
@@ -131,6 +131,8 @@ public:
     void set_appearance(const Appearance &a);
     const Color &get_color(ColorP colorp) const;
 
+    bool touchpad_pan = false;
+
     bool smooth_zoom = true;
     bool snap_to_targets = true;
 
--- a/src/canvas/pan.cpp
+++ b/src/canvas/pan.cpp
@@ -63,9 +63,8 @@ void CanvasGL::pan_drag_move(GdkEventScroll *scroll_event)
 {
     gdouble dx, dy;
     gdk_event_get_scroll_deltas((GdkEvent *)scroll_event, &dx, &dy);
-
-    offset.x += dx * 50;
-    offset.y += dy * 50;
+    offset.x -= dx * 50;
+    offset.y -= dy * 50;
     update_viewmat();
     queue_draw();
 }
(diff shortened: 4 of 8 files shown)


carrotIndustries:
Seems like `gdk_device_get_n_axes` isn't a good way to tell a multitouchpad apart from a regular one. On my machine with a non-multitouch touchpad, it returns 4.

How does the inversion interact with the natural scrolling setting of the desktop environment? My touchpad is configured for edge scrolling and by default, moving the finger to the left moves the content to the right. With the inversion, touchpad scrolling also behaves this way.

anuejn:
> Seems like gdk_device_get_n_axes isn't a good way to tell a multitouchpad apart from a regular one. On my machine with a non-multitouch touchpad, it returns 4.

ah huh, that's bad :(

> How does the inversion interact with the natural scrolling setting of the desktop environment? My touchpad is configured for edge scrolling and by default, moving the finger to the left moves the content to the right. With the inversion, touchpad scrolling also behaves this way.

Natural scrolling changes the direction of panning. Good catch.

I see mostly two ways to proceed:

1. Introduce settings for inversion + Multitouchpad pan Zoom (not optimal)
2. Use the gtk gesture recognition and somehow disable the threshold for geasture recognition in gtk for touchpads (I havent found any way to do that; more work)

carrotIndustries:
>   Introduce settings for inversion + Multitouchpad pan Zoom (not optimal)

On that note, I noticed that the trackpoint panning is 'natural' by default, i.e. the opposite of other applications, so we should probably fix this as well. To me at least, being consistent with GtkScrolledWindow is good enough without any special logic to invert the scrolling direction on our side.

KiCad actually has exactly the same setting that you have in mind (Use touchpad to pan), but implemented in a not so good way as it doesn't discern mice from touchpad and uses clunky non-smooth scrolling. We can do better :)

After having wrote this comment, I'm fine with a setting that'll turn touchpad scroll events into pan. Unsurprisingly, this setting should go into `ZoomPreferences`.

anuejn:
Ok, I implemented what you asked for :)

carrotIndustries:
LGTM feature-wise. Some minor nitpicks:

 - To reduce coupling, don't read the preferences in the canvas itself. Instead, add a public member variable and set it in `preferences_provider_attach_canvas` and `ImpBase::apply_preferences`
 - With the new setting, we should probably rename the Zoom tab in the preferences to `Zoom & Pan`
 - Run clang-format to fix style

anuejn:
Can you re-review?

carrotIndustries:
There are merge conflicts, please rebase onto latest master.

src/canvas/canvas_gl.cpp
```diff
@@ -8,6 +8,9 @@
 #include <glm/gtc/type_ptr.hpp>
 #include "board/board_layers.hpp"
 #include "bitmap_font_util.hpp"
+#include "preferences/preferences_provider.hpp"
+#include "preferences/preferences.hpp"
```
carrotIndustries: we don't need these anymore

src/imp/imp_board.cpp
```diff
@@ -230,6 +230,7 @@ void ImpBoard::apply_preferences()
     }
     canvas->set_highlight_on_top(preferences.board.highlight_on_top);
     canvas->show_text_in_tracks = preferences.board.show_text_in_tracks;
+    canvas->touchpad_pan = preferences.zoom.touchpad_pan;
```
carrotIndustries: this should go to `ImpBase::apply_preferences` as we want this to apply too all editors

src/canvas/canvas_gl.cpp
```diff
@@ -617,6 +617,11 @@ void CanvasGL::set_appearance(const Appearance &a)
     queue_draw();
 }
 
+void CanvasGL::set_touchpad_pan(bool touchpad_pan)
```
carrotIndustries: use a different argument name to avoid shadowing warnings, i.e. `a_touchpad_pan`

anuejn: btw: do you prefer a setter or acessing the field directly?

carrotIndustries: Depends. Since this field has no particular restrictions on its value, accessing the field without a setter is fine from my perspective. If you prefer a setter, I'm fine with that as well.

src/preferences/preferences_util.cpp
```diff
@@ -14,10 +14,12 @@ void preferences_provider_attach_canvas(CanvasGL *ca, bool layer)
         a = &PreferencesProvider::get_prefs().canvas_non_layer.appearance;
     }
     ca->set_appearance(*a);
+    ca->touchpad_pan = PreferencesProvider::get_prefs().zoom.touchpad_pan;
     PreferencesProvider::get().signal_changed().connect(sigc::track_obj(
             [ca, a] {
                 ca->set_appearance(*a);
                 ca->smooth_zoom = PreferencesProvider::get_prefs().zoom.smooth_zoom_2d;
```
anuejn: shouldnt this also be set outside the change listener?

src/canvas/canvas_gl.cpp
```diff
@@ -617,6 +617,7 @@ void CanvasGL::set_appearance(const Appearance &a)
     queue_draw();
 }
 
+
```
carrotIndustries: unnecessary newline

src/preferences/preferences_util.cpp
carrotIndustries: good catch, missed that

carrotIndustries:
see 1e6539a0202cd260f9560d75f979e892bb23cd2b for `smooth_zoom`

PR #367, opened 2020-04-03
footprint generator: add basic silkscreen, reference designators and set parameter class of courtyard

m-byte:
These changes apply to the IPC footprint generator and add the following:
1. A simple rectangular silkscreen is automatically added to components generated by the footprint generator. Pads are substracted from the silkscreen outline to make it look nicer.
2. Reference designators are added on assembly layer and silk screen. The reference designator on assembly layer is scaled to fit the package outline.
3. The courtyard polygon is assigned the proper parameter class.

With these changes, packages generated automatically pass the "package checks" rules.

Still missing:
- Reference designator on silkscreen is not positioned correctly if pads are extending past the bottom edge of the package.
- There is no pin 1 marker of any kind.

### DIFF
--- a/src/imp/footprint_generator/footag/display.cpp
+++ b/src/imp/footprint_generator/footag/display.cpp
@@ -1,4 +1,5 @@
 #include <regex>
+#include <iomanip>
 #include "display.hpp"
 #include "widgets/preview_canvas.hpp"
 #include "widgets/spin_button_dim.hpp"
@@ -348,6 +349,11 @@ static const Padstack *getpadstack(Pool &pool, enum footag_padstack stack)
     return ps;
 }
 
+static short compare_range(double a, double b, double c)
+{
+    return c >= a && c <= b ? 0 : c < a ? -1 : 1;
+}
+
 static void make_rlimit_rect(Polygon &poly, const struct footag_rlimit *r, enum BoardLayers::Layer layer)
 {
     poly.vertices.clear();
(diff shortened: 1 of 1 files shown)


src/imp/footprint_generator/footag/display.cpp
```diff
@@ -356,6 +362,117 @@ static void make_rlimit_rect(Polygon &poly, const struct footag_rlimit *r, enum
...
+    }
+
+    for (std::list<std::array<Coordi, 2>>::iterator it=lines.begin(); it != lines.end(); ++it) {
+        points.push_back(it->at(0));
+        points.push_back(it->at(1));
+    }
+    points.sort();
+    points.unique();
+
+    for (std::list<Coordi>::iterator point=points.begin(); point != points.end(); ++point) {
```
carrotIndustries, line 454: Use ranged for loops if applicable: `for(const auto &it: points) {...`

src/imp/footprint_generator/footag/display.cpp
```diff
@@ -356,6 +362,117 @@ static void make_rlimit_rect(Polygon &poly, const struct footag_rlimit *r, enum
...
+    points.unique();
+
+    for (auto point = points.begin(); point != points.end(); ++point) {
+        auto uu = UUID::random();
+        auto junction = &pkg->junctions.emplace(uu, uu).first->second;
+        junction->position = *point;
+        junctions.push_back(junction);
+    }
+
+    for (auto coords = lines.begin(); coords != lines.end(); ++coords) {
```
carrotIndustries, line 461: Any reason for not using ranged for loops here?

m-byte: Sorry about that. I just noticed that I forgot cleaning up my code in some places before pushing (in part because circleci reminded me). Didn't even expect anyone to take a look at this so quickly, so I only read your comment after pushing a slightly cleaner version. For most of the loops, I will change to ranged ones. The only one that I likely won't be able to change is the one in line 415 as I have to modify the iterator when deleting lines.

Btw: that is some impressively fast response time! I'm used to pull request sitting around for weeks until someone actually takes a look at them :D

m-byte: Code is now cleaned up and tested. Also, I changed all for loops except for the one in 415 to ranged for loops.

endofexclusive:
I really like the idea of automatic generation of silkscreen.

However, I wonder if the IPC-7351B footprint generator is the right place for this. IPC-7351B is a land pattern standard. It does explicitly not specify silkscreens and the standard document does not even contain the words _silkscreen_ or _reference designator_. (My understanding of the IPC committee stance on this is that board artwork (silkscreen etc) is an organizational policy which should be adapted to the design requirements and fabrication capabilities at hand. Or something like that.)

Instead, how about a dedicated tool for automatic generation of silkscreen? Something like the `Generate courtyard` tool:
- It could be used on any footprint, manually entered or generated.
- Could have a set of knobs and buttons for selecting styles, line widths, expansion from pad and body, font height etc.
- Could have a catalog of template styles. (Horizon convention, KiCad convention, custom conventions, etc.).
- Possibility to select a script for custom artwork which gets the current footprint features in a dict?

If the PC-7351B generator shall generate silkscreens, then I highly recommend to make the silkscreen an opt-in with a check-mark. The silkscreen should also show up in the preview when the check-mark is activated.

carrotIndustries:
> If the PC-7351B generator shall generate silkscreens, then I highly recommend to make the silkscreen an opt-in with a check-mark. The silkscreen should also show up in the preview when the check-mark is activated.

Thanks for chiming in, since you did the IPC footprint generator, it's up to you to decide what we want in there and what not.

m-byte:
> Instead, how about a dedicated tool for automatic generation of silkscreen? Something like the `Generate courtyard` tool:
> 
> * It could be used on any footprint, manually entered or generated.
> * Could have a set of knobs and buttons for selecting styles, line widths, expansion from pad and body, font height etc.
> * Could have a catalog of template styles. (Horizon convention, KiCad convention, custom conventions, etc.).
> * Possibility to select a script for custom artwork which gets the current footprint features in a dict?

Thanks for the suggestion. This is probably the best option - as you said, it would be great for all kinds of footprints.

When I started writing this code, my intention was to add the parameter class "courtyard" to courtyards generated by the IPC generator. Once I dug into the code, however, I started thinking "what if it could also do x?" and that's how this code came to be.

With this in mind, I will close this pull request as is and first open a new one for the "courtyard" parameter class in the IPC generator and then another one once I'm done restructuring my code to make this a tool instead.

Two more questions:
What about the reference designator on assembly layer? I'm not sure if it makes sense to have that as a separate tool. On the one hand, it doesn't have anything to do with the silkscreen. On the other hand I find it a little annoying to have to use multiple tools on every footprint... Or should this be part of the IPC generator, @endofexclusive?

@carrotIndustries as I haven't yet dug far enough into the code, yet... Is there a way for tools to automatically run other tools? E.g. to have a more generic "generate (missing) footprint features" that would then allow you to call "generate courtyard", "generate silkscreen", ... with the settings you last used (best with the option to select what you want).

carrotIndustries:
> Is there a way for tools to automatically run other tools?

There is `ToolResponse::next` to invoke another tool right at the end of a tool, but it's probably not what you're looking for.

The common way how dissimilar tools share code is through `ToolHelper...` classes, such as `ToolHelperMapSynbol`.

Another way is to implement similar tools in one class and then use `tool_id` to figure out what's to be done. i.e. `ToolAddPart` is used for both `ADD_PART` and `ADD_COMPONENT`.

endofexclusive:
If you feel that implementing a tool for silkscreen+assembly is something you want to try, then I definitely think that is the right thing to do. It would be a very useful tool to invoke after manually entering a manufacturer recommended footprint from a datasheet for example. The assembly layer generation can be part of the same tool.

I don't see a big issue from a workflow perspective to generate the silkscreen+assembly in a dedicated tool, since the effort of invoking the tool would anyway be marginal compared to the effort of entering the lead+component parameters in the IPC generator or copying from a datasheet illustration.


If you are not going for a dedicated tool however, then a solution is to have the silkscreen+assembly generation as part of the IPC generator. It is convenient since there is quite some information about the footprint features available from the footag library. In that case, I recommend:
- Next to the "Auto fit", there can be checkboxes for "Silkscreen" and "Assembly"
- The `calc()` can take the checkboxe value into account for adding the silkscreen and assembly layers.
- For preview, the filtering in `display()` can be changed to also have `visible` on the silkscreen and assembly layers.

m-byte:
@endofexclusive As you said, having it as a separate tool is great for manually entered components, so that is what I went for. Since your component generator already generates a polygon in package and this can reasonably be expected to be done by the user for manually entered components, I use this as a reference. Next, I will try to add things like text, pin 1 markers, ...

PR #368, opened 2020-04-04
footprint generator: add courtyard parameter class to polygon

m-byte:
Previously part of https://github.com/horizon-eda/horizon/pull/367, now only the courtyard parameter is changed in IPC generator


endofexclusive:
Great! I tried the patch and it works very well.

PR #370, opened 2020-04-05
new tool: generate silkscreen

m-byte:
As suggested by @endofexclusive, I moved the silkscreen generator from https://github.com/horizon-eda/horizon/pull/367 into a tool. Right now it creates an outline by offsetting the package polygon and subtracting all pads from it. I will next turn this into a tool that allows for dynamic changes of the offset. This tool does not yet take care of adding a reference designator.


m-byte:
With commit 4a3d4bf, the tool is interactive and both offsets (pad/polygon) can be adjusted. Also, if there are multiple package polygons, the tool selects the first one and the user can change this selection.

fruchti:
This is a great addition! A minor nitpick: The default pad/package expansion should probably be 0.2 mm, in accordance with the [current pool convention](https://github.com/horizon-eda/horizon-pool-convention).

m-byte:
Thanks for the reminder. 0.15mm was in there because I used the same value for line width and expansion in the very first version. I'll update it once I get time to implement proper handling of settings.

m-byte:
@fruchti here you go. Settings are now properly implemented and default is 0.2 mm.

fruchti:
Works great, thank you!

endofexclusive:
This is cool! Really nice how the silkscreen can snake around corners etc.

I observed one small oddity while using the tool on the default SO footprint in the IPC-7351:
silk0
In the screenshot the lower horizontal line extends farther to the right than to the left. It is visible only for particular combinations of outline and pad expansion. Here it was 0.7 mm pad and 0.5 mm outline expansion. The upper horizontal line looks as expected though.

The tool could of course be extended with lots of style options and parameters. But I think it already does a great job and improves the package editor workflow already.

carrotIndustries:
Very cool to see someone who isn't me implement a nontrivial tool without any hand holding!

The current way of adjusting the expansion parameters is a bit too clunky IMO, and I think there's a  better way: For not all that long we have a way for tools to show a non modal dialog to adjust more complex parameters on the fly, see the renumber pads tool for an example. In our case a window with just two spinbuttons (use `SpinButtonDim`) to adjust the spacing parameters is probably all we need.

When it comes to pin 1 markings, keep in mind that there are two styles for pin 1 markings:

1. Omit one of the lines in the corner if the package doesn't have pins that extend beyond the package's body (QFN)
2. Extend the line if pins extend beyond the package's body (SOIC, QFP)

m-byte:
@endofexclusive That's a bug that seems to be caused by a rounding error. I checked the points that I'm feeding into ClipperLib and they are correct. So the root cause seems to be within Clipper. Until someone figures out what is going wrong, I added an extra .000001 mm offset to the pad expansion.

@carrotIndustries pin 1 markers are not implemented at the moment. For now, the user would have to take care of that manually. But I believe that is rather trivial. Especially on packages with curved packages, I find drawing a correct outline with cutouts for pads most challenging (and annoying).

Right now, I'm looking into the modal dialog you mentioned and will probably implement it relatively soon.

Longer term, I'm planning to add other styles (e.g. corners only), pin 1 markers (I'm planning to have the options none, extended/omitted line, arrow, dot) and silkscreen text. For these, a dialog is likely the best interface. But since the current status is enough to satisfy my immediate need, I first wanted to get some feedback on the current status to make sure I don't waste time on features no one will ever use.

carrotIndustries:
> But since the current status is enough to satisfy my immediate need, I first wanted to get some feedback on the current status to make sure I don't waste time on features no one will ever use.

No worries, with the **non** modal dialog in place, this PR is ready to be merged.

m-byte:
Sometimes my hands are faster than my head :smile:
As requested, here's the *non* modal dialog.

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+                if (!pp) {
+                    pp = &it.second;
+                }
+                else {
+                    ret = false;
+                }
+            }
+        }
+    }
+
```
carrotIndustries, line 59: Exit here if `pp == nullptr` to avoid crashing.

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+        path_pkg.emplace_back(ClipperLib::IntPoint(it.position.x, it.position.y));
+    }
+    return ret;
+}
+
+void ToolGenerateSilkscreen::clear_silkscreen()
+{
+    auto pkg = doc.k->get_package();
+    std::list<UUID> silklines;
+    for (const auto &it : pkg->lines) {
```
carrotIndustries, line 75: use `map_erase_if` from `util.h` to turn this into a one-liner

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    for (const auto &it : pkg->pads) {
+        for (const auto &it_poly : it.second.padstack.polygons) {
+            if (it_poly.second.layer == BoardLayers::TOP_COPPER) {
+                ClipperLib::Path pad_path;
+                auto polygon = it_poly.second.remove_arcs();
+                for (const auto &vertex : polygon.vertices) {
+                    auto pos = it.second.placement.transform(vertex.position);
+                    pad_path.emplace_back(ClipperLib::IntPoint(pos.x, pos.y));
```
carrotIndustries, line 116: `pad_path.emplace_back(pos.x, pos.y); ` is all that's needed

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+        }
+        else {
+            return ToolResponse();
+        }
+    }
+    else if (args.type == ToolEventType::CLICK && args.button == 1 && args.target.layer == BoardLayers::TOP_PACKAGE
+        && (args.target.type == ObjectType::POLYGON_EDGE || args.target.type == ObjectType::POLYGON_VERTEX)) {
+        selection.clear();
+        selection.emplace(
+                SelectableRef(args.target.path.at(0), args.target.type, args.target.vertex, args.target.layer));
```
carrotIndustries, line 187: `SelectableRef(..)` can be omitted here

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+        select_polygon();
+    }
+    else if (first_update) {
+        first_update = false;
```
carrotIndustries, line 191: What's the purpose of `first_update`?

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+        return ToolResponse();
+    }
+
+    clear_silkscreen();
+
+    ClipperLib::ClipperOffset ofs_pads;
+    ClipperLib::Paths pads_expanded;
+    ofs_pads.AddPaths(pads, ClipperLib::jtMiter, ClipperLib::etClosedPolygon);
+    /* additional .000001_mm offset helps with rounding errors */
+    ofs_pads.Execute(pads_expanded, settings.expand_pad + .075_mm + .000001_mm);
```
carrotIndustries, line 216: Introduce a constant `line_width` somewhere or make it part of the settings to avoid magic numbers.

src/dialogs/generate_silkscreen_window.cpp
```diff
@@ -0,0 +1,53 @@
+#include "generate_silkscreen_window.hpp"
+//#include "pool/package.hpp"
+//#include "widgets/spin_button_dim.hpp"
```
carrotIndustries, line 3: remove commented includes

src/dialogs/generate_silkscreen_window.cpp
```diff
@@ -0,0 +1,53 @@
...
+    grid->set_column_spacing(10);
+    grid->set_row_spacing(10);
+    grid->set_margin_bottom(20);
+    grid->set_margin_top(20);
+    grid->set_margin_end(20);
+    grid->set_margin_start(20);
+
+    int top = 0;
+
+    const int width_chars = 9;
```
carrotIndustries, line 25: not needed since we don't show/hide rows on the fly

src/dialogs/generate_silkscreen_window.cpp
```diff
@@ -0,0 +1,53 @@
...
+    sp_silk = Gtk::manage(new SpinButtonDim);
+    sp_silk->set_range(-1_mm, 5_mm);
+    bind_widget(sp_silk, settings->expand_silk);
+    sp_silk->signal_value_changed().connect(sigc::mem_fun(*this, &GenerateSilkscreenWindow::update));
+    grid_attach_label_and_widget(grid, "Outline offset", sp_silk, top)->set_width_chars(width_chars);
+
+    sp_pad = Gtk::manage(new SpinButtonDim);
+    sp_pad->set_range(-1_mm, 5_mm);
```
carrotIndustries, line 33: Do we really need negative spacings for silkscreen and package?

src/dialogs/generate_silkscreen_window.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "util/gtk_util.hpp"
+#include "util/util.hpp"
+#include <math.h>
+
+namespace horizon {
+GenerateSilkscreenWindow::GenerateSilkscreenWindow(Gtk::Window *parent, ImpInterface *intf, ToolSettings *stg)
+    : ToolWindow(parent, intf)
+{
+    set_title("Generate silkscreen");
+    settings = (ToolGenerateSilkscreen::Settings *)stg;
```
carrotIndustries, line 13: use `settings(dynamic_cast<ToolGenerateSilkscreen::Settings*>(stg))` in the member initializer list instaed and perhaps make the argument and the member a reference

src/dialogs/generate_silkscreen_window.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+class GenerateSilkscreenWindow : public ToolWindow {
+public:
+    GenerateSilkscreenWindow(Gtk::Window *parent, class ImpInterface *intf, class ToolSettings *stg);
+
+    void update();
+
+private:
+    class Package *pkg = nullptr;
+    std::set<class Pad *> pads;
```
carrotIndustries, line 20: copy/paste leftovers most likely

carrotIndustries:
Works perfectly! Some nitpicking:

 - Style check failed
 - Is persistently saving the settings for expansion a good idea? We have some defaults that should be used in the majority of cases. The current situation makes it quite hard to get back to the defaults. So either provide a way to reset the settings back to default or don't store them persistently in the first place.
 - Changing the layer display settings doesn't let the layer box on the left side doesn't pick up changes made directly to the canvas. The proper way to do this would be to manipulate the layer box through the `ImpInterface`. Unfortunately that infrastructure isn't there yet, but isn't too hard to add if needed.
- See code comments

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,264 @@
...
+    for (const auto &it : pkg->pads) {
+        for (const auto &it_poly : it.second.padstack.polygons) {
+            if (it_poly.second.layer == BoardLayers::TOP_COPPER) {
+                ClipperLib::Path pad_path;
+                auto polygon = it_poly.second.remove_arcs();
+                for (const auto &vertex : polygon.vertices) {
+                    auto pos = it.second.placement.transform(vertex.position);
+                    pad_path.emplace_back(ClipperLib::IntPoint(pos.x, pos.y));
+                }
+                pads.emplace_back(pad_path);
```
carrotIndustries, line 118: Use `pads.emplace_back(std::move(pad_path));` to explicitly allow for moving the contents of `pad_path`.

m-byte:
Thanks for the thorough review!

- Saving settings: I guess that comes down to the user. Yes, the defaults reflect the style guide for the horizon pool. But what if users have their own corporate guidelines regarding style that are different? Then, they will have to change all the settings every time. But I get your point and will add a "defaults" button (if there are better suggestions for the name, let me know).
- Style check: That's odd. I ran `clang-format-9 -style=file -i` on the file before pushing.
- Proper layer display manipulation: I don't think that should be part of this pull request, so I believe leaving it as is until a proper interface is there should be okay. Do you agree?

carrotIndustries:
> Proper layer display manipulation: I don't think that should be part of this pull request, so I believe leaving it as is until a proper interface is there should be okay. Do you agree?

Okay, I'll try to add the proper interface once the PR is merged.

src/dialogs/generate_silkscreen_window.cpp
m-byte: Not me personally, that's why I didn't write `(-5_mm, 5_mm)`, but maybe someone has a use for it. I'll change it to `(0, 5_mm)`, though. Then if someone needs it, they can always modify this.

src/core/tools/tool_generate_silkscreen.cpp
m-byte: I used it to run update (for drawing) the first time after starting the tool. However, I realized now, that I can get the same result using `ToolEventType::NONE` which I pass when calling it from my begin function.

endofexclusive:
The dialog is a nice improvement.

Some comments (minor):

1. The tool changes layer.
I don't see it as a big problem and maybe it is intended. But it behaves different compared to the tools `Place pad` and `Generate courtyard` which do not change the layer as it finishes.
2. Return/enter could finish the tool, or progress to the next widget in the non-modal.
There is code for `ToolEventType::KEY` but it seems to not reach the tool from the window (?).
3. Escape key to cancel the tool?

m-byte:
> 1. The tool changes layer.
This is intentional, so the user can see the silkscreen. Of course - like for the package layer - I could store the current working layer and then return to it after the silkscreen is done. Is that preferred?
> 2. Return/enter could finish the tool, or progress to the next widget in the non-modal.
> 3. Escape key to cancel the tool?
The code for both escape and enter is in the tool, so it would work if you select the main window. I haven't yet implemented it for the dialog, but I also feel like this should be done on a higher level (i.e. as part of `ToolWindow`) so that these basic key bindings are the same for all tools using windows.

endofexclusive:
> This is intentional, so the user can see the silkscreen. Of course - like for the package layer - I could store the current working layer and then return to it after the silkscreen is done. Is that preferred?

No real preference from my side. I am fine with how it works now.

> The code for both escape and enter is in the tool, so it would work if you select the main window. I haven't yet implemented it for the dialog, but I also feel like this should be done on a higher level (i.e. as part of `ToolWindow`) so that these basic key bindings are the same for all tools using windows.

Aaah, that explains it. Indeed the keys work when the window does not have focus. I previously had the dialog focused while trying out the hot keys. The window manager here (i3) always gives the window focus as it opens (did not try with other managers). To be continued in #375 then :-).

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,263 @@
...
+        auto canvas = imp->get_canvas();
+        auto ld = canvas->get_layer_display(BoardLayers::TOP_PACKAGE);
+        package_visible = ld.visible;
+        ld.visible = true;
+        canvas->set_layer_display(BoardLayers::TOP_PACKAGE, ld);
+    }
+
+    win = imp->dialogs.show_generate_silkscreen_window(&settings);
+
+    return ToolResponse::change_layer(BoardLayers::TOP_SILKSCREEN);
```
carrotIndustries, line 161: `ToolResponse::change_layer` might removed sooner or later (only used by the legacy router), use `imp->set_work_layer`

src/core/tools/tool_generate_silkscreen.cpp
```diff
@@ -0,0 +1,263 @@
...
+            if (data->event == ToolDataWindow::Event::CLOSE) {
+                restore_package_visibility();
+                return ToolResponse::revert();
+            }
+            else if (data->event == ToolDataWindow::Event::OK) {
+                restore_package_visibility();
+                return ToolResponse::commit();
+            }
+            else if (data->event == ToolDataWindow::Event::UPDATE) {
+            }
```
carrotIndustries, line 177: add a comment such as `// it's okay, go ahead` to signify that this `if` is empty on purpose

carrotIndustries:
Sorry for not catching anything at first sight, these should be trivial to fix. The layout of the "Default" button still needs some improvement as the window is a bit left-heavy right now:
 - Attach the button to the grid, below the entries and make it span both columns
 - That way, we have more space in the button, e.g. "Restore defaults"
 - set the grid's halign to CENTER

src/core/tools/tool_generate_silkscreen.cpp
m-byte: Okay. Wasn't sure, which one would be preferred.

src/core/tools/tool_generate_silkscreen.cpp
m-byte: Since I wansn't happy with this logic (it's easy to miss one of the else cases and the corresponding `ToolResponse()`), I refactored `update` so that there is now a function `redraw_silkscreen`. That takes care of this.

carrotIndustries:
29b9174e1fe76eff21dc92c0c840d69f72316015 fixes the layer display inconsistencies

m-byte:
Great! Thank you!

PR #375, opened 2020-04-07
tool window: keyboard control

m-byte:
As mentioned in https://github.com/horizon-eda/horizon/pull/370, it would be nice to use the tool window with keyboard keys. Right now, only the escape key is enabled, but this should be a good starting point for discussion.
An extension of this would be for the enter/return key to first step through all input boxes and then confirm the dialog. (Maybe it would be better to press return twice without any other keys in between to avoid accidental confirmation when editing the last input)


carrotIndustries:
We have `install_esc_to_close(Gtk::Window &win)` in `gtk_util.hpp` to reduce the boilerplate.

m-byte:
Do you have any idea how I can find out when enter is pressed while focus is on a Gtk::Entry?
I tried `signal_key_press_event().connect()`, but return/enter seems to be filtered out.

carrotIndustries:
`signal_activate` it is. See https://github.com/horizon-eda/horizon/blob/master/src/dialogs/ask_datum.cpp#L36 for an example of how this is to be used.

pwnorbitals:
Has this PR been abandonned @m-byte ?

m-byte:
For the time being... Quite busy with family and regular work at the
moment, so I'm not expecting to have any time to work on this until early
2022. Feel free to close or modify this as you see fit.

Am Fr., 30. Apr. 2021 um 10:09 Uhr schrieb [IPSA] Chris de Claverie <
***@***.***>:

PR #376, opened 2020-04-09
add support for file:// links

m-byte:
This code adds support for `file://` links. Both relative and absolute paths are supported. Relative paths are relative to the directory of the currently opened file. So for example `file://schematic.pdf` opens `schematic.pdf` from the project directory, while `file:///user/foo/schematic.pdf` would open a file from the home directory of user `foo`.

### DIFF
--- a/src/imp/imp.cpp
+++ b/src/imp/imp.cpp
@@ -923,15 +923,23 @@ std::string ImpBase::get_hud_text(std::set<SelectableRef> &sel)
     if (sel_count_type(sel, ObjectType::TEXT) == 1) {
         const auto text = core->get_text(sel_find_one(sel, ObjectType::TEXT).uuid);
         const auto txt = Glib::ustring(text->text);
-        auto regex = Glib::Regex::create(R"(https?:\/\/([\w\.-]+)(\/\S+)?)");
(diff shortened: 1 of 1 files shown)


m-byte:
Please note that I don't know how well this works on Windows. Likely, the code for relative paths will have to be modified.

src/imp/imp.cpp
```diff
@@ -923,15 +923,23 @@ std::string ImpBase::get_hud_text(std::set<SelectableRef> &sel)
     if (sel_count_type(sel, ObjectType::TEXT) == 1) {
         const auto text = core->get_text(sel_find_one(sel, ObjectType::TEXT).uuid);
         const auto txt = Glib::ustring(text->text);
-        auto regex = Glib::Regex::create(R"(https?:\/\/([\w\.-]+)(\/\S+)?)");
+        auto regex = Glib::Regex::create(R"((https?:\/\/|file:\/\/\/?)([\w\.-]+)(\/\S+)?)");
```
carrotIndustries, line 926: I'd prefer separating `file` and `http` links to avoid duplicating regexes.

carrotIndustries:
Links to URLs seem to be broken on windows at the moment due to https://gitlab.gnome.org/GNOME/glib/-/issues/2088 or something else.

PR #381, opened 2020-04-19
fonts: add mapping for plus minus sign

m-byte:
I needed a plus minus sign in my schematic and noticed that it was not mapped to Hershey.


carrotIndustries:
Glyph 2233 more closely matches the + glyph.

m-byte:
Ok, changed it

PR #383, opened 2020-05-04
Redraw color selector when loading color scheme.

jmgao:
Previously, when loading a color scheme or restoring to default, the
list of colors would keep their previous values until a redraw was
forced by selecting another color or modifying the currently selected
color.

### DIFF
--- a/src/pool-prj-mgr/preferences_window_canvas.cpp
+++ b/src/pool-prj-mgr/preferences_window_canvas.cpp
@@ -70,7 +70,6 @@ public:
 protected:
     virtual std::string get_color_name() = 0;
     Gtk::DrawingArea *colorbox = nullptr;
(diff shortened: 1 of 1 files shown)


carrotIndustries:
Out of interest, on which platform did you encounter this? On X11, it works as intended whereas I'm able to reproduce the issue on wayland/gnome.

jmgao:
This was on X11 with i3wm. I'd guess that on wayland, the back window isn't requested to repaint the region obscured by the dropdown menu because it's being drawn to an off-screen buffer. No idea why it repros on i3, though. AFAIK I'm not using any sort of compositor.

carrotIndustries:
> This was on X11 with i3wm

That's odd, as this is exactly my setup as well...

PR #387, opened 2020-05-07
Sheet References

m-byte:
This PR enables sheet references in the schematic.
My use case for this are block diagrams. There it is very helpful to reference another page. Since these references are dynamically generated using the UUID of the respective sheet, they are automatically updated whenever pages are put in a new order.
Additionally, sheet references are shown in the HUD.

### DIFF
--- a/src/imp/imp_schematic.cpp
+++ b/src/imp/imp_schematic.cpp
@@ -521,8 +521,9 @@ void ImpSchematic::construct()
                     if (sch->sheets.count(uuid)) {
                         sheet_box->select_sheet(uuid);
                     }
+                    return true;
                 }
-                return true;
+                return false;
             },
             false);
 
--- a/src/imp/imp_schematic.cpp
+++ b/src/imp/imp_schematic.cpp
@@ -516,7 +516,6 @@ void ImpSchematic::construct()
             [this](const std::string &url) {
                 if (url.find("s:") == 0) {
                     std::string sheet = url.substr(2);
-                    std::cout << "sheet " << sheet << std::endl;
                     auto uuid = UUID(sheet);
                     auto sch = core_schematic.get_schematic();
                     if (sch->sheets.count(uuid)) {
@@ -600,11 +599,8 @@ std::string ImpSchematic::get_hud_text(std::set<SelectableRef> &sel)
     if (sel_count_type(sel, ObjectType::TEXT) == 1) {
         const auto text = core->get_text(sel_find_one(sel, ObjectType::TEXT).uuid);
         const auto txt = Glib::ustring(text->text);
-        auto regex = Glib::Regex::create(
-                R"(\$sheetref_([a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}))",
-                Glib::RegexCompileFlags::REGEX_CASELESS);
         Glib::MatchInfo ma;
-        if (regex->match(txt, ma)) {
+        if (core_schematic.get_schematic()->get_sheetref_regex()->match(txt, ma)) {
             s += "\n\n<b>Text with sheet references</b>\n";
             do {
                 auto uuid = ma.fetch(1);
@@ -613,8 +609,9 @@ std::string ImpSchematic::get_hud_text(std::set<SelectableRef> &sel)
                     s += make_link_markup(url, core_schematic.get_schematic()->sheets.at(UUID(uuid)).name);
                 }
                 else {
-                    s += make_link_markup(url, "Unknown Sheet");
+                    s += "Unknown Sheet";
                 }
+                s += "\n";
             } while (ma.next());
             sel_erase_type(sel, ObjectType::TEXT);
         }
--- a/src/imp/imp_schematic.cpp
+++ b/src/imp/imp_schematic.cpp
@@ -512,6 +512,21 @@ void ImpSchematic::construct()
         return false;
     });
 
+    main_window->signal_activate_hud_link().connect(
+            [this](const std::string &url) {
+                if (url.find("s:") == 0) {
+                    std::string sheet = url.substr(2);
+                    std::cout << "sheet " << sheet << std::endl;
+                    auto uuid = UUID(sheet);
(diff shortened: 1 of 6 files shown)


src/imp/imp_schematic.cpp
```diff
@@ -574,6 +589,29 @@ std::string ImpSchematic::get_hud_text(std::set<SelectableRef> &sel)
...
+        if (regex->match(txt, ma)) {
+            s += "\n\n<b>Text with sheet references</b>\n";
+            do {
+                auto uuid = ma.fetch(1);
+                std::string url = "s:" + uuid;
+                if (core_schematic.get_schematic()->sheets.count(UUID(uuid))) {
+                    s += make_link_markup(url, core_schematic.get_schematic()->sheets.at(UUID(uuid)).name);
+                }
+                else {
+                    s += make_link_markup(url, "Unknown Sheet");
```
carrotIndustries, line 609: why make a link if the sheet is not found?

src/imp/imp_schematic.cpp
```diff
@@ -574,6 +589,29 @@ std::string ImpSchematic::get_hud_text(std::set<SelectableRef> &sel)
...
+                }
+            } while (ma.next());
```
carrotIndustries, line 611: add newline after link so that we get one link per line

src/schematic/schematic.cpp
```diff
@@ -795,6 +796,34 @@ void Schematic::expand(bool careful)
...
+
+    if (!careful) {
+        auto regex = Glib::Regex::create(
+                R"(\$sheetref_([a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}))",
+                Glib::RegexCompileFlags::REGEX_CASELESS | Glib::RegexCompileFlags::REGEX_OPTIMIZE);
+        for (auto &it_sheet : sheets) {
+            Sheet &sheet = it_sheet.second;
+            for (auto &it_text : sheet.texts) {
+                auto &text = it_text.second;
+                auto txt = text.text;
```
carrotIndustries, line 808: use `Glib::ustring txt = text.text;` to prevent memory corruption. This is a bit unfortunate as `regex->match` expects a `Glib::ustring&` that gets constructed temporarily from our `std::string` but then goes out of scope, but is referenced afterwards. As far as I can tell, this is fixed in glibmm master, but may need further investigation...

src/schematic/schematic.cpp
```diff
@@ -795,6 +796,34 @@ void Schematic::expand(bool careful)
             }
         }
     }
+
+    if (!careful) {
+        auto regex = Glib::Regex::create(
```
carrotIndustries, line 801: In terms of style, I'd prefer making it `$sheetref:` rather than `$sheetref_`. We should probably make the regex a static method of schematic to avoid code duplication, like so

```
Glib::Refptr<Glib::Regex> Schematic::get_sheetref_regex() {
 static auto regex = ....;
 return regex;
}

**Comment by carrotIndustries** [comment]
Cool, but wouldn't it make more sense to show the sheet name rather than the sheet's number?

**Comment by m-byte** [comment]
Thanks for the review.
Yes and no... In my case, I wanted to replicate the style that is used by off-sheet labels in my block diagram. It would be possible to have two versions of this: one for sheet number and one for sheet name. (e.g. `$sheetref:` and `$sheetname:`)

File: src/imp/imp_schematic.cpp (hunk 1 above)
Comment (m-byte) [reply src/imp/imp_schematic.cpp:609]: It was mostly due to copy paste... I'll change it to just read `unknown sheet`.

File: src/schematic/schematic.cpp [hunk 5]
```diff
@@ -970,6 +966,15 @@ std::map<UUIDPath<2>, std::string> Schematic::get_unplaced_gates() const
     return unplaced;
 }
 
+Glib::RefPtr<Glib::Regex> Schematic::get_sheetref_regex()
+{
+    if (!sheetref_regex)
```
carrotIndustries, line 971: just make `sheetref_regex` static and let the compiler take care of it: https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

m-byte: Interesting, didn't know that was possible.

src/schematic/schematic.hpp
```diff
@@ -81,6 +81,8 @@ class Schematic {
 
     std::map<UUIDPath<2>, std::string> get_unplaced_gates() const;
 
+    Glib::RefPtr<Glib::Regex> get_sheetref_regex();
```
carrotIndustries, line 84: we can make this `static`

m-byte:
Updated now

src/imp/imp_schematic.cpp
```diff
@@ -512,6 +512,20 @@ void ImpSchematic::construct()
...
+            [this](const std::string &url) {
+                if (url.find("s:") == 0) {
+                    std::string sheet = url.substr(2);
+                    auto uuid = UUID(sheet);
+                    auto sch = core_schematic.get_schematic();
+                    if (sch->sheets.count(uuid)) {
+                        sheet_box->select_sheet(uuid);
+                    }
+                }
+                return true;
```
carrotIndustries, line 525: this needs to return `false` to pass control up to the default handlers

src/imp/imp_schematic.cpp
```diff
@@ -512,6 +512,20 @@ void ImpSchematic::construct()
...
 
+    main_window->signal_activate_hud_link().connect(
+            [this](const std::string &url) {
+                if (url.find("s:") == 0) {
+                    std::string sheet = url.substr(2);
+                    auto uuid = UUID(sheet);
+                    auto sch = core_schematic.get_schematic();
+                    if (sch->sheets.count(uuid)) {
+                        sheet_box->select_sheet(uuid);
+                    }
```
carrotIndustries, line 523: `return true;` here to stop default handlers

m-byte: You are correct. Somehow while working on this, `return true;` was accidentally moved outside the if and then the `return false;` didn't make sense anymore. Fixed now

PR #388, opened 2020-05-07
Fix copying frame of current sheet to new sheet

m-byte:
Currently, when creating a new sheet, instead of copying the frame reference of the current frame, the frame contents are copied. This results in frames with the contents of the one on the previous sheet. Additionally, users have to manually select the frame for every sheet for proper operation.

By copying pool_frame instead of frame, this can be fixed.


carrotIndustries:
Good catch!

PR #395, opened 2020-05-13
Disable outline creation in pdf export for podofo version 0.9.5

m-byte:
There is a bug that existed in podofo from revision 1775 to 1826. Because of this, only version 0.9.5 should be affected. With this bug it is not possible to create a PdfOutlineItem.

Since e.g. Ubuntu 18.04 LTS ships with this version podofo, it is not possible to export PDFs from horizon on Ubuntu 18.04 with libpodofo from the official repositories. The best remedy for this is to skip creating an outline when linked against podofo 0.9.5 since all the other steps run fine.

### DIFF
--- a/src/export_pdf/export_pdf.cpp
+++ b/src/export_pdf/export_pdf.cpp
@@ -40,10 +40,12 @@ void export_pdf(const class Schematic &sch, const class PDFExportSettings &setti
 
     auto font = document.CreateFont("Helvetica");
 
+#if PODOFO_VERSION_MAJOR != 0 || PODOFO_VERSION_MINOR != 9 || PODOFO_VERSION_PATCH != 5
     auto outlines = document.GetOutlines();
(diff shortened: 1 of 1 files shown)


carrotIndustries:
> With this bug it is not possible to create a PdfOutlineItem.

Does it crash? Why can't we leave it as-is?

m-byte:
> > With this bug it is not possible to create a PdfOutlineItem.
> 
> Does it crash? Why can't we leave it as-is?

Yes. It fails with ePdfError_OutlineItemAlreadyPresent

m-byte:
Since I find it easier to read changes introduced by a commit on github than on sourceforge, here's a link to the fix that arrived shortly after 0.9.5 was released: https://github.com/svn2github/podofo/commit/d66fca5d6b2243d8bae47b966f91a0355b262a7e#diff-12df260f321b420bbbc1915ca3ba08c1

Issue #119, opened 2018-02-21
[UI] IMP : add a toolbar for selecting tools


jeanthom:
Currently most tools are only available through keyboard shortcuts. Having a toolbar with all the tools (draw net, place text, etc.) could improve the learning curve of the schematic and board editor.

carrotIndustries:
> Currently most tools are only available through keyboard shortcuts

There's always the spacebar menu providing a searchable list of all tools with their shortcuts right next to them.

TBH, I don't like toolbars at all. After some time they usually end up looking like this: https://i.stack.imgur.com/tPIf4.png

At first sight, many of the icons could be hieroglyphs as well without reducing their meaningfulness, so the long-term user interaction goes like this: The user hovers over all icons to learn from their tooltip what clicking on them will do. After some time, she remembers "clicking on this white-blue colored blob will recalculate all cooper planes".  By that time she could have remembered "press q to recalculate all copper planes" as well.

The spacebar menu itself may be a bit hard to discover indeed, but once the user found out about it, all tools (soon actions) are easily discoverable.

jeanthom:
Well, maybe we shouldn't include all the tools available in the toolbar, but at least the core features.

I don't know much about Cadence software (last time I used it was like a year ago, and for a very short period of time). On other EDA software I've used (Eagle, KiCad), the main issue with toolbars is that everything is crammed in (both tools selection & tools properties). That makes the whole toolbar crowded and indigestible to a newbie.

I'd like to hear feedback from other users about shortcuts and spacebar menu.

atoav:
I must say I am on the same side as @carrotIndustries on this topic, I would avoid a toolbar with a ton of symbols – I think promoting the spacebar menu and well thought shortcuts makes more sense.

If there really should be a mouse-only way of selecting commands or tools, I could imagine a well organized nested menu like for example `Edit → Copy` and `Draw → Line` etc. But I dont really think that is necessary. I drew ~30 packages with the current version and the Spacebar Menu did it's job quite well for me. Nested menu doesn't necessarily mean dropdown menu, I will come back to this below.

[...]

PR #396, opened 2020-05-14
Spacebar menu hint

carrotIndustries:
As discussed in #119, I've had the idea to show some kind of hint to make newcomers aware of the spacebar menu. Current design:

2020-05-14-223120_538x482_scrot

Things left to figure out:
 - When to show this hint? I've had in mind to show this if the spacebar menu hasn't been opened within 30(?) seconds after opening an editor
 - When not to show this hint? Set a a flag that the spacebar menu has been opened at least once and not show it?

Any thoughts?


fruchti:
I like the design! IMHO, it would feel less unexpected if it was shown immediately on the first start, though. With the time delay it might happen that a new user opens an example schematic, zooms and pans around and has the tip window pop up unrequested, maybe even closing it by accident without getting a chance to read the text.

endofexclusive:
Nice.
- when to show
  - Agree about showing it immediately on the first start.
  - When pressing F1?

- when not to show
  - Anytime after the first start.

The spacebar menu is powerful and effective. I have found myself hitting space in all kinds of other applications to invoke a command, just to realize that I was not in Horizon.

About the hit design. How about adding an image of an actual spacebar? Now there is the light bulb icon so the user might try to find that one in the GUI. The text "Press space" could be "Press the space bar" to prevent mixing it up with the schematic work space for example.

carrotIndustries:
I wanted to give the user some time to explore the application on their own since users have developed some kind of blindness towards anything that pops up right after opening an application...

carrotIndustries:
How about adding another button to the action bar that opens the menu?
2020-12-31-151830_482x506_scrot

carrotIndustries:
> How about adding another button to the action bar that opens the menu?

One downside of this approach is that some tools such as "move" aren't really useful from this menu as they depend on where the cursor is on tool activation. To me at least that's not that big a deal as the main purpose of the menu in this place is to inform users about pressing the spacebar. The info text on top of the menu is supposed to include a link to the preferences to remove the menu button.

carrotIndustries:
> One downside of this approach is that some tools such as "move" aren't really useful from this menu as they depend on where the cursor is on tool activation.

This could be mitigated by hiding tools that depend on the cursor position from that menu, how about this?

bob-u:
I would prefer to see all the available options at all times.  Maybe a notice somewhere within the popup window could inform a user that selection is required for function to work?  Or, gray our functions that are not available.

carrotIndustries:
So this is what we've got now:
2021-07-09-231437_214x158_scrot

Might not be as pretty as some of the other suggestions, but should get the job done.

PR #405, opened 2020-05-31
Show pin name on pad in board editor

endofexclusive:
This commit adds a new toggle switch when a package is selected in the
board editor.

Setting this toggle in the "on" position forces pin name to be
displayed on every pad. Setting this toggle to "off" gives the previous
behavior of showing the assigned net name if available.

Motivation:
1. It is useful to be able to distinguish between proper power pads and
pads which have been assigned to a power net. For example configuration pins.

2. Adds a visual cue when doing fan-out of I/O pads which do not yet
have a net name assigned.

The solution is simple and has the following limitations:
- The switch enters the *off* position when doing history undo.
- Takes vertical space from the right hand side panel.

### DIFF
--- a/src/board/board.cpp
+++ b/src/board/board.cpp
@@ -816,6 +816,8 @@ void Board::expand_packages()
                 }
                 else {
                     it_pad.second.net = nullptr;
+                }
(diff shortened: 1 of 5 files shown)


bob-u:
Great idea.  I have been missing this feature also.

PR #413, opened 2020-06-10
Add distraction free mode

rroohhh:
Currently this just hides the left panel and the property panel.

Open questions:
- should more things be hidden? I think the way it currently works is fine.
- should there be some kind of indication that one is in distraction free mode? I don't think its necessary, but could help people that enter it on accident.

### DIFF
--- a/src/imp/action.hpp
+++ b/src/imp/action.hpp
@@ -89,6 +89,7 @@ enum class ActionID {
     GEN_FAB_OUTPUT,
     FOOTPRINT_GENERATOR,
     SET_GRID_ORIGIN,
+    DISTRACTION_FREE,
 };
 
 using ActionToolID = std::pair<ActionID, ToolID>;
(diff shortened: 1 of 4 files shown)


rroohhh:
Well other than maybe some feedback on the open questions, this is ready to merge I think

carrotIndustries:
Works great. Some kind of visual feedback (perhaps in the status bar) would be nice, but since this mode isn't persistent across application restarts, confused people can just reopen the editor, so we can merge it as-is in that regard.

Apart from that, the action should be called 'Toggle distraction free mode'.

rroohhh:
Ok, made it part of the `view_option_menu` and added it to the view hints.

carrotIndustries:
The place in the view options menu isn't perfect as all other options only affect what happens in the canvas itself, but I couldn't think of a better place either.

PR #451, opened 2020-07-07
Allow hiding of DNP components in 3D view

m-byte:
This adds an extra switch to 3D view settings that can be used to show or hide DNP components. Right now, it is called "DNP 3D Models" and off by defaults. Not very happy with the name, but I couldn't think of anything shorter that brings the point across...

### DIFF
--- a/src/canvas3d/canvas3d_base.cpp
+++ b/src/canvas3d/canvas3d_base.cpp
@@ -294,11 +294,11 @@ void Canvas3DBase::prepare_packages()
         return;
     package_transform_idxs.clear();
     package_transforms.clear();
-    std::map<std::string, std::set<const BoardPackage *>> pkg_map;
+    std::map<std::pair<std::string, bool>, std::set<const BoardPackage *>> pkg_map;
     for (const auto &it : brd->packages) {
         auto model = it.second.package.get_model(it.second.model);
         if (model)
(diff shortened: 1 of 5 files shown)


carrotIndustries:
LGTM, even though it took a me longer than I'd have liked to to recall how exactly the 3D model loading in the Canvas3D works.

In terms of UI, how about replacing both switches with a single 3-button ganged switch (use radio buttons, set draw-indicator to false and pack them in a box with style class linked)? We'd than have these three options
 - None
 - Placed
 - All

m-byte:
How about a dropdown instead? That would make it easy to add e.g. assembly variants at some point without significant UI changes.
I don't like naming the intermediate option "Placed" since that could be confusing to some users. Altium, for example, also words it the other way around (https://www.altium.com/documentation/altium-designer/workspacemanager-dlg-variantoptionformvariant-options-ad). How about "Hide unpopulated"?

carrotIndustries:
> How about a dropdown instead?

I'm fine with that as well, but these ganged switches are less fiddly to operate. And the user can see all available options at a glance.

> I don't like naming the intermediate option "Placed" since that could be confusing to some users

I'm not particularly happy with it either, but it's short enough to look okay in a ganged switch.

m-byte:
Updated as per your suggestion

PR #484, opened 2020-08-03
Fix DRC for plated through holes

m-byte:
Current master does not actually check distance between copper and plated holes.
With this fix, DRC ignores the layer for plated through holes. This is necessary because they are not located on the actual copper layers. The fixed version can catch boards that have incorrectly entered components and can prevent e.g. power planes being shorted in the final board.

### DIFF
--- a/src/board/board_rules_check.cpp
+++ b/src/board/board_rules_check.cpp
@@ -258,6 +258,17 @@ RulesCheckResult BoardRules::check_clearance_copper(const Board *brd, RulesCheck
                         patch_pairs.emplace(k);
                     }
                 }
+                else if (layer == it.first.layer && it.first.net != it_other.first.net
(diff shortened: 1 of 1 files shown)


carrotIndustries:
Good catch! Just to be sure, this is referring to the PTH "barrel"? In which situations do we have PTH barrels without a pad?

m-byte:
Yes, that's what I'm referring to. Might only be the case for incorrectly entered components (i.e. entered only for dual layer but used on multi layer), but in our case this check would have made the difference between a batch of unusable PCBs and usable ones.
From my understanding, DRC should be able to catch things you can do wrong, not only what you are expected to do wrong.

carrotIndustries:
Hm, seems like the plane update needs to fixed as well: f3e813e0b74bae763aa12179f36dd0640f720f76

PR #485, opened 2020-08-03
Fix appveyor by installing keys of new MinGW packagers

m-byte:
See https://www.msys2.org/news/#2020-06-29-new-packagers


carrotIndustries:
Thank you so much! I was hoping that this issue would fix itself by appveyor upgrading their msys2 installation or some other project coming up with a fix that's easy to paste, but a mergable PR is even better!

m-byte:
I mostly did this because the little red x next to the recent PRs were annoying me :D

PR #487, opened 2020-08-03
Makefile: remove gold linker, fix indentation.

ericonr:
Gold linker brings little benefit, isn't available or fully compliant
for some architectures, and can have issues with libraries located
outside /usr/lib.

Rust tried enabling it, but had to revert: https://github.com/rust-lang/rust/pull/30913

When packaging Horizon for Void Linux, I had to remove  `-fuse-ld=gold` from the Makefile, since it isn't available for our cross compilation setup.

It was enabled some years ago for no reason that I could find.

This way, if someone wants it, it can still be passed on the environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -669,19 +669,21 @@ endif
 
 
 ifeq ($(OS),Windows_NT)
-    LDFLAGS += -lrpcrt4
-    DEFINES += -DWIN32_UUID
-    LDFLAGS_GUI = -Wl,-subsystem,windows
+	LDFLAGS += -lrpcrt4
+	DEFINES += -DWIN32_UUID
+	LDFLAGS_GUI = -Wl,-subsystem,windows
 else
-    UNAME := $(shell uname)
-    ifeq ($(UNAME), FreeBSD)
-        CXXFLAGS += -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR
-    endif
-    ifeq ($(UNAME), Darwin)
-        # do nothing on mac os
-    else
-        LDFLAGS += -fuse-ld=gold
-    endif
+	UNAME := $(shell uname)
+	ifeq ($(UNAME), FreeBSD)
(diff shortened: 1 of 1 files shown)


carrotIndustries:
> It was enabled some years ago for no reason that I could find.

The gold linker is about twice as fast as the regular one. Linking horizon-imp takes 10s with the gold linker, 20s with the regular one. 

> This way, if someone wants it, it can still be passed on the environment.

Why not make it the other way round and keep gold the default and add an option to use the non-gold ld?

ericonr:
> Why not make it the other way round and keep gold the default and add an option to use the non-gold ld?

Sure! Could it be something like

```
GOLD = -fuse-ld=gold
[...]
LDFLAGS += $(GOLD)
```

and a user can use `make GOLD=` ?

carrotIndustries:
> Sure! Could it be something like

Yes, go ahead.

ericonr:
Done!

Makefile
```diff
@@ -669,19 +669,21 @@ endif
...
-    endif
+	UNAME := $(shell uname)
+	ifeq ($(UNAME), FreeBSD)
+		CXXFLAGS += -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR
+	endif
+	ifeq ($(UNAME), Darwin)
+		# do nothing on mac os
+	else
+		# allow ld.gold to be overriden by command line
+		GOLD = -fuse-ld=gold
```
carrotIndustries, line 684: shouln't this be `GOLD ?= -fuse-ld=gold` to allow for overriding?

ericonr: `?=` will take the value from the environment, if it exists, so with it you could do `GOLD= make`. But calling `make GOLD=` will fully replace `GOLD` in the makefile, no matter how it's defined.

I didn't think of making it controllable by environment variables, but can switch to it.

PR #522, opened 2020-10-13
Updated stm32_to_json script

katyo:
Because original script looks outdated, I reimplemented it.

### DIFF
--- a/scripts/stm32_to_json.py
+++ b/scripts/stm32_to_json.py
@@ -1,30 +1,178 @@
 import xml.etree.ElementTree as ET
-import sys
+import argparse
+import uuid
+import os
 import json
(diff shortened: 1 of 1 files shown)


carrotIndustries:
The original script outputs json that's to be imported into the part wizard for manual adjustments. Either keep that option in your script or add your script as a new one.

katyo:
> The original script outputs json that's to be imported into the part wizard for manual adjustments. Either keep that option in your script or add your script as a new one.

Hmm... It seems I failed to get it working.

katyo:
@carrotIndustries 
May be it will be appropriate to move this PR to __horizon-pool__ repo?

PR #526, opened 2020-10-26
Add holes rendering to PDF export

katyo:
KiCAD has "Drill marks" option which looks like so:
kicad_drill_marks

Because currently Horizon missing similar features, I trying to add it.
image

### DIFF
--- a/src/board/board.cpp
+++ b/src/board/board.cpp
@@ -435,8 +435,9 @@ void Board::update_pdf_export_settings(PDFExportSettings &settings)
 {
     auto layers_from_board = get_layers();
     // remove layers not on board
-    map_erase_if(settings.layers,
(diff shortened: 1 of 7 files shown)


katyo:
Currently I added pseudo-layer for holes to settings:
pdf_export_settings_hl
This solution may looks not so proper but it adds some flexibility for export.

## Export examples

Holes with actual sizes:
holes_actual_size

Holes with specified sizes:
holes_specified_size

Holes with actual sizes (outline):
holes_outline

Holes over mask:
holes_over_mask

katyo:
@carrotIndustries 
I made holes diameter spinner to be insensitive when actual size mode is on.

carrotIndustries:
Board `app_version` should be bumped. See https://blog.horizon-eda.org/progress/2020/10/29/progress-2020-09-10.html#version-info-in-files for details

PR #540, opened 2020-12-31
circle support DXF import, curved airwire

chocopi:
I have added circle support for DXF import with 2 arcs.
circle_support_DXF

And curved airwires and different colored airwires for power net. I hope there is a dedicated color setting for them(currently Router airwire for power net).
curved_airwire
I hope there is an option for a user to select between the curved or straight airwire.


carrotIndustries:
Please split this into individual PRs. I also just added circle/arc drawing routines to the canvas that make use of the recently introduced shaders.

bob-u:
Curved wires seem to be the solution for overlapping wires.  Pretty cool!

PR #541, opened 2021-01-02
Makefile: improve parallelism.

jmgao:
Previously, the $(shell pkg-config ...) invocation was called for every
single file, because CXXFLAGS was a recursively expanded variable.


carrotIndustries:
Cool, this reduced the time for a cached rebuild with 100% hit rate from 40s to 20s!

PR #542, opened 2021-01-02
[WIP] smooth normals in 3D preview

carrotIndustries:
To improve the appearance of models in the 3D preview, I had a go at implementing smooth normals. Normals are computed using opencascade:
```
  //! Compute node normals for face triangulation
  //! as mean normal of surrounding triangles
  Standard_EXPORT static void ComputeNormals (const Handle(Poly_Triangulation)& Tri);
```

While this works okay-ish most of the time, it causes artifacts  when used on closed (is that how it's called?) faces such a cylinder's barrel: 
2021-01-02-152416_301x401_scrot

Looking at the normals, we can see that we have two at the vertex where the face closes:
2021-01-02-152502_860x639_scrot

@atoav Is there a recommended way to avoid artifacts like this?


atoav:
Hey, not sure. But having a look at Blender's [edge split modifier](https://docs.blender.org/manual/en/latest/modeling/modifiers/generate/edge_split.html) might be useful: Sourcode is [here](https://github.com/blender/blender/blob/594f47ecd2d5367ca936cf6fc6ec8168c2b360d0/source/blender/modifiers/intern/MOD_edgesplit.c)

carrotIndustries:
Apart from that, the edges where curved and flat faces meet look a bit odd: 
2021-01-02-190927_1516x567_scrot

atoav:
Not sure why it is like that there, but usually such a thing has to do with a sudden change in curvature. Curvature is 1/radius so when you are going from a circle to a flat surface you will go from e.g. radius 1 (curvature 1) to radius infinity (curvature 0) instantly, that can look odd in certain lighting conditions even with real objects which is why such surface transitions are usually avoided in cars bodies etc.

Not sure if this is the whole problem there however. If the last face normal of the fillet points in the same direction as the top surface it should just look the same, unless there is a hard edge in the model (or duplicate faces etc)

atoav:
In Blender there are three types of normals:
- Face normals
- Vertex Normals
- Split Normals

As you can see the "properly" smoothed mesh has all of these 3 aligned, while they differ for the unsmoothed mesh:
split normals  
split normals 2

carrotIndustries:
> if the last face normal of the fillet points in the same direction as the top surface it should just look the same, unless there is a hard edge in the model 

My suspicion is that this isn't the case and a fillet looks like this in 2D so we get the hard edge. The red section is a different face than the black straight sections.
2021-01-02-202339_352x341_scrot

OTOH, Kicad looks roughly the same, so this is proably okay:
2021-01-02-192159_794x526_scrot

Kicad also shows the artifact at the polygon barrel: 
2021-01-02-202936_562x542_scrot

atoav:
Probably an issue with the mesh then I guess? Could have a look at it

carrotIndustries:
> Probably an issue with the mesh then I guess? Could have a look at it

Is there an easy format I could export the mesh to? The mesh data structures looks like this: https://github.com/horizon-eda/horizon/blob/c539a06a6f9974403f06f0ebcd86602450d85b59/src/util/step_importer.hpp#L34-L40

Unfortunately, I believe that there's not much we can do about this since OpenCASCADE for mesh generation isn't exactly easy to work with.

atoav:
OBJ would be straightforward I think?

carrotIndustries:
There you go: [obj.zip](https://github.com/horizon-eda/horizon/files/5760914/obj.zip) The faces are separated by an empty line. In the `-fixed` cylinder, normals at coincident vertices of the same face are averaged.

Looking at the normals, we do indeed seem to have the problem I described:
2021-01-02-224642_494x490_scrot

As far as I can tell the flipped normals at the concave fillet aren't an issue for the flat shading used in the 3D preview.

atoav:
Ok found the issue the cornering fillet vertecies (see the ones I selected) seem to be doubled. In blender this can be fixed by a "merge by distance" operation, which also makes the transition look smooth smooth. This is a quite common issue with some generators that don't produce clean connected meshes:

2021-01-02 22_59_46-Blender

atoav:
That means these 3 vertecies should (in a clean mesh) all become one (and the same is true for all other corners etc:  
2021-01-02 23_04_05-Blender

carrotIndustries:
> Edit: a good idea is probably to implement a "merge by distance" function and just run it on the whole mesh.

Hm, seems like that's not that easy since since each face is processed independently of other/neighboring faces. Unless there's a way to get opencascade to do this for us or there's an easy to use library, I don't see this happening in the near future.

Anyhow, I think that the smooth normals, even though not perfect, still are an improvement over the flat normals. What do you think?

atoav:
I think this is acceptable : )

PR #544, opened 2021-01-08
Makefile: only run pkg-config for python when needed.

jmgao:


carrotIndustries:
Doesn't look like so: https://github.com/horizon-eda/horizon/pull/544/checks?check_run_id=1668602415#step:6:4

carrotIndustries:
obsolete since we now have meson

PR #561, opened 2021-02-05
Fix draw connection line on pins on package origin.

jmgao:
When a pin is directly on the origin of a package (e.g. pin 1 on
every generic pin-header), the draw connection tool wouldn't work on
said pin because it'd receive the BoardPackage instead.

It seems like the ideal fix would be for tools to declare what object
types they're expecting to receive for an action, and have the tool
infrastructure do the disambiguation, but for now, when this happens,
scan the package for a pad on the origin to use instead.

### DIFF
--- a/src/core/tools/tool_draw_connection_line.cpp
+++ b/src/core/tools/tool_draw_connection_line.cpp
@@ -45,11 +45,40 @@ ToolResponse ToolDrawConnectionLine::update(const ToolArgs &args)
         return ToolResponse();
     }
     else if (args.type == ToolEventType::ACTION) {
+        const ObjectType target_type = args.target.type;
         switch (args.action) {
         case InToolActionID::LMB:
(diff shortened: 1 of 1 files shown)


src/core/tools/tool_draw_connection_line.cpp
```diff
@@ -45,11 +45,40 @@ ToolResponse ToolDrawConnectionLine::update(const ToolArgs &args)
         return ToolResponse();
     }
     else if (args.type == ToolEventType::ACTION) {
+        ObjectType target_type = args.target.type;
```
carrotIndustries, line 48: can be const

carrotIndustries:
LGTM apart from the const nitpicking.

src/core/tools/tool_draw_connection_line.cpp
jmgao: Done.

PR #573, opened 2021-02-17
Add customization of silkscreen color.

jmgao:


carrotIndustries:
LGTM, we might consider bumping the board's version since the on-disk format has changed: https://github.com/horizon-eda/horizon/blob/e25f4c00cbdd6529c1e9a2ca7e4e64dc57722992/src/board/board.cpp#L37 When doing so, don't forget to add this to the changelog: https://github.com/horizon-eda/horizon/blob/master/scripts/app_versions.yml

jmgao:
> LGTM, we might consider bumping the board's version since the on-disk format has changed:
> 
> https://github.com/horizon-eda/horizon/blob/e25f4c00cbdd6529c1e9a2ca7e4e64dc57722992/src/board/board.cpp#L37
> 
> When doing so, don't forget to add this to the changelog: https://github.com/horizon-eda/horizon/blob/master/scripts/app_versions.yml

Done.

PR #574, opened 2021-02-17
step export: don't export non-populated models.

jmgao:


carrotIndustries (changes requested):
`n_pkg` should be set accordingly, i.e. `n_pkg  = count_if(... it->component->populate)`

jmgao:
> `n_pkg` should be set accordingly, i.e. `n_pkg = count_if(... it->component->populate)`

Done (by not adding them to the intermediate vector and setting n_pkg to that vector's size, instead)

carrotIndustries (approved):

PR #604, opened 2021-06-09
Set default corner radius in dual inline footprint generator

frmdstryr:
Currently if using the rounded rect padstack it shows a missing parameter when using the footprint generator.  Might be a better way to do this?  Should I add this to the quad/inline generators as well?

Peek 2021-06-09 19-11

### DIFF
--- a/src/imp/footprint_generator/footprint_generator_base.hpp
+++ b/src/imp/footprint_generator/footprint_generator_base.hpp
@@ -7,9 +7,6 @@
 #include <set>
 
 namespace horizon {
-class Pad;
-class Padstack;
-
 class FootprintGeneratorBase : public Gtk::Box {
 public:
     FootprintGeneratorBase(const char *resource, class IDocumentPackage &c);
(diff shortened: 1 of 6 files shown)


src/imp/footprint_generator/footprint_generator_base.hpp
```diff
@@ -24,5 +26,8 @@ class FootprintGeneratorBase : public Gtk::Box {
     Gtk::Box *box_top = nullptr;
     class IDocumentPackage &core;
     class Package &package;
+
+    virtual void update_pad_parameters(const Padstack *padstack, Pad &pad, const int64_t pad_width,
```
carrotIndustries, line 30: `padstack` can be a reference as it must be non-null. You can forward-declare `Pad`

src/imp/footprint_generator/footprint_generator_base.hpp
```diff
@@ -5,7 +5,11 @@
 #include <array>
 #include <gtkmm.h>
 #include <set>
+
 namespace horizon {
+class Pad;
```
carrotIndustries, line 10: You can do the forward declaration in the method declaration: `class Pad &pad`

src/imp/footprint_generator/footprint_generator_base.hpp
```diff
@@ -24,5 +28,8 @@ class FootprintGeneratorBase : public Gtk::Box {
     Gtk::Box *box_top = nullptr;
     class IDocumentPackage &core;
     class Package &package;
+
+    virtual void update_pad_parameters(const Padstack &padstack, Pad &pad, const int64_t pad_width,
```
carrotIndustries, line 32: Doesn't need to be virtual.

frmdstryr:
Thanks for the help :+1:

PR #607, opened 2021-06-18
Hud show track spacing

frmdstryr:
When two parallel tracks are selected show the spacing between them.

image

### DIFF
--- a/src/board/track.cpp
+++ b/src/board/track.cpp
@@ -216,4 +216,15 @@ bool Track::coord_on_line(const Coordi &p) const
     }
     return false;
 }
+
+bool Track::is_parallel_to(const Track &other) const
+{
+    const auto zero = Coordi();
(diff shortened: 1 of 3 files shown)


src/board/track.cpp
```diff
@@ -216,4 +216,15 @@ bool Track::coord_on_line(const Coordi &p) const
     }
     return false;
 }
+
+bool Track::is_parallel_to(const Track &other) const
+{
+    const auto zero = Coordi();
+    const auto v1 = to.get_position() - from.get_position();
+    const auto v2 = other.to.get_position() - other.from.get_position();
+    if (v1 == zero or v2 == zero)
```
carrotIndustries, line 225:
```suggestion
    if (v1 == zero || v2 == zero)
```

To be consistent with everything else.

carrotIndustries:
Excellent implementation! Thanks for this PR and the other one.

PR #609, opened 2021-07-01
Use arcs when exporting to pdf

frmdstryr:
Instead of converting circles/arcs to polylines this exports them as circles/arcs. This needs more testing but it is working on the SMD rounded rectangular pads and vias. 

1. Should I add an option to toggle it on/off?
2. How can I test the arcs?

After putting them in inkscape and doing a path union:

Before

image

After

image

### DIFF
--- a/src/export_pdf/canvas_pdf.cpp
+++ b/src/export_pdf/canvas_pdf.cpp
@@ -9,6 +9,7 @@
 
 namespace horizon {
 
+
 CanvasPDF::CanvasPDF(PoDoFo::PdfPainterMM &p, PoDoFo::PdfFont &f, const PDFExportSettings &s)
     : Canvas::Canvas(), painter(p), font(f), settings(s), metrics(font.GetFontMetrics())
 {
@@ -155,13 +156,14 @@ void CanvasPDF::img_polygon(const Polygon &ipoly, bool tr)
             }
             else if (last->type == Polygon::Vertex::Type::ARC) {
                 // Finish arc
-                Coordi start = last->position;
-                Coordi c = last->arc_center;
+                Coordd start = last->position;
+                Coordd c = last->arc_center;
+
                 if (tr) {
                     c = transform.transform(c);
                     start = transform.transform(start);
                 }
-                const auto r = ((Coordd) start - (Coordd) c).mag();
+                const auto r = (start - c).mag();
                 if (start == p) {
                     painter.Circle(to_pt(c.x), to_pt(c.y), to_pt(r));
                 }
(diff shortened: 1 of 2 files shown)


carrotIndustries:
> Should I add an option to toggle it on/off?

I don't see any good reason for not doing this as long as it works reliably.

> How can I test the arcs?

What do you mean by this? I drew a rectangle with one rounded corner and it doesn't look like a rectangle at all in the PDF.

frmdstryr:
>  as long as it works reliably.

That was my concern, might not be an issue, idk.

> What do you mean by this? I drew a rectangle with one rounded corner and it doesn't look like a rectangle at all in the PDF.

I just haven't used arcs anywhere yet.  Glad it worked. 

Also, it doesn't do holes, not sure if that should be changed?

carrotIndustries:
> I just haven't used arcs anywhere yet. Glad it worked.

I'm afraid, it didn't: 
2021-07-01-234733_952x483_scrot

> Also, it doesn't do holes, not sure if that should be changed?

Makes sense as well.

frmdstryr:
Perhaps this can be cleaned up, arcs are enough to make someone go insane... but hopefully the logic is right now. 

I could not figure out how to get podofo's ArcTo to work and [Arc](https://github.com/svn2github/podofo/blob/a7fd39e6446cb56aafade49219e956f60fd3b755/src/doc/PdfPainter.cpp#L1590) only does clockwise arcs which seems to break the close path command so it has to be manually closed when ending on a lineto.

carrotIndustries:
> I could not figure out how to get podofo's ArcTo to work and Arc

I haven't hat a closer look at the Podofo docs, but for other aspects, it's been helpful to consult the PDF spec itself as podofo is a rather thin abstraction on top of PDF.

frmdstryr:
It looks like the pdf paths only support bezier curves.  I spent more time playing with ArcTo and I give up again, I cannot get it to make any sane arc. 

I decided to just implement an arc function...  took a while but I finally got it to work on a bunch of stuff I added to a project and hopefully it is clean enough..

image

frmdstryr:
Bump. Is there any other way you'd prefer this be implemented or should I just drop it?

carrotIndustries:
Thanks for the nudge, I've been busy with implementing hierarchical schematics the past days. With a bit more work, I think we could adapt the arc drawing code to also draw line arcs as actual arcs. This would necessitate the addition of something like `img_arc` to `Canvas` so that derived canvases such as the PDF one get to know about arcs. By having a default implementation that draws line segments, other canvases can be left as-is.

PR #616, opened 2021-10-13
new layer display style: dotted

rroohhh:
This adds the `DOTTED` layer style. The circles are arranged on a hexagonal grid (I also tried a square grid, but that looked a lot worse). 

Furthermore, because this is (as the hatched style) discard based and the circles are quite small the missing anti-aliasing is very apparent. I don't really mind it, but what do you think?

20211013_20h36m09s_grim


carrotIndustries:
Saving a board with the layer display set to dotted throws an exception since the new enumerator is missing from the lut:
https://github.com/horizon-eda/horizon/blob/02edcc308d096721537df8d011b78c3b0f7a1fba/src/widgets/layer_box.cpp#L324-L329

The only concern I have is that there's now one more option to click through in the layer box. Adding a context menu to the display mode selector could fix this.

Issue #619, opened 2021-10-29
Feature Request: Restrict 45 degrees for polygon 


kamtar:
I do like using planes a lot in my designs but it's hard to keep them looking neat without 45-degree snap-on/restriction as in Altium.

I'm thinking about writing it myself but I have yet to go through the codebase so I just wanted to mention it, and ask if I'm not missing something and if you think that it's reasonably easy to implement.

carrotIndustries:
That's actually rather straightforward to add. The relevant line is https://github.com/horizon-eda/horizon/blob/5194b3aafc7a63a29d883e4fbc43e2efbceeb5ed/src/core/tools/tool_draw_polygon.cpp#L91 It calculates the position of the vertex being placed based on the cursor position and the last vertex position.

As the coordinate calculation is factored out into the `ToolHelperRestrict` class, it might make sense to add the 45° snap in there. That way the move and draw line tool could also make use of it. I wouldn't have it selectable by the usual <kbd>/</kbd> keybinding though as 4 items are a bit too many to cycle through. Instead, I suggest adding a new in-tool action to select the 45° mode.

kamtar:
Thanks for the pointers.
Are you interested in including it in the main branch? Should I bother doing a pull request?

carrotIndustries:
>  Are you interested in including it in the main branch? Should I bother doing a pull request?

Yes, I think that's a feature worth having.

kamtar:
I went with a default key "q" for now.
It gets toggled between on and off by q, also if it's enabled and you press / for the restrict mode it cancels it and jumps to ARB.

It copies one axis and calculates the other one based on which axis is the longest. I was thinking about creating more modes and having auto, X angle, Y angle but I don't see the use in it yet. 

It jumps to angles based on what's closer to the cursor. Meaning if you are doing a 45-degree line and you move your cursor only on one axis it will jump back on the axis rather soon. Any thoughts on this? I was thinking if I shouldn't implement some hysteresis.

I added it to a polygon and line tool, I was looking at adding it to a move tool too but that looks much more involved so I shall leave it for now.

carrotIndustries:
Even if it's not completely done yet, it's best to open a PR.

PR #620, opened 2021-10-30
45 degree polygon restrict

kamtar:
As discussed in #619

### DIFF
--- a/src/core/tools/tool_helper_restrict.cpp
+++ b/src/core/tools/tool_helper_restrict.cpp
@@ -54,7 +54,7 @@ std::string ToolHelperRestrict::restrict_mode_to_string() const
         return "Y only";
         break;
     case RestrictMode::DEG45:
-        return "45 degrees only";
+        return "45° only";
         break;
     }
     return "";
@@ -78,8 +78,8 @@ Coordi ToolHelperRestrict::get_coord_restrict(const Coordi &old, const Coordi &c
 Coordi ToolHelperRestrict::find_45deg_coord(const Coordi &old, const Coordi &cur) const
 {    
     Coordi result;
-    Coordi vect = cur - old;
-    Coordi vect_abs = {std::abs(vect.x), std::abs(vect.y)};
+    const Coordi vect = cur - old;
+    const Coordi vect_abs = {std::abs(vect.x), std::abs(vect.y)};
 
     if(vect_abs.x >= vect_abs.y){
         result.x = cur.x;
--- a/src/core/tools/tool_helper_restrict.cpp
+++ b/src/core/tools/tool_helper_restrict.cpp
@@ -34,9 +34,10 @@ void ToolHelperRestrict::cycle_restrict_mode_xy()
 
 void ToolHelperRestrict::toogle_45_degrees_mode()
 {
-    if(restrict_mode == RestrictMode::DEG45){
+    if (restrict_mode == RestrictMode::DEG45) {
         restrict_mode = RestrictMode::ARB;
-    }else{
+    }
+    else {
         restrict_mode = RestrictMode::DEG45;
     }
 }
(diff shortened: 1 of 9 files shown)


kamtar:
I see that the stylecheck generated a diff with fixed issues. Is that something I need to apply manually and commit or is that automated later in the process?

carrotIndustries:
Run soemthing similar to `git diff master --name-only | grep '\.cpp\|\.hpp' | xargs clang-format -style=file -i` to run the autoformatter.

src/core/tools/tool_helper_restrict.cpp
```diff
@@ -41,6 +53,9 @@ std::string ToolHelperRestrict::restrict_mode_to_string() const
     case RestrictMode::Y:
         return "Y only";
         break;
+    case RestrictMode::DEG45:
+        return "45 degrees only";
```
carrotIndustries, line 57: Space in the tool bar is quite limited, I'd use the ° sign:
```suggestion
        return "45° only";
```

src/imp/in_tool_action_catalog.cpp
```diff
@@ -82,6 +82,7 @@ const std::map<InToolActionID, InToolActionCatalogItem> in_tool_action_catalog =
          {"Flip direction", ToolID::ADD_VERTEX, InToolActionCatalogItem::FLAGS_DEFAULT}},
 
         {InToolActionID::ARC_MODE, {"arc mode", ToolID::DRAW_ARC, InToolActionCatalogItem::FLAGS_DEFAULT}},
+        {InToolActionID::TOGGLE_DEG45_RESTRICT, {"toggle 45 degree restrict", ToolID::NONE, InToolActionCatalogItem::FLAGS_DEFAULT}},
```
carrotIndustries, line 85: Same as above:
```suggestion
        {InToolActionID::TOGGLE_DEG45_RESTRICT, {"toggle 45° restrict", ToolID::NONE, InToolActionCatalogItem::FLAGS_DEFAULT}},
```

src/core/tools/tool_helper_restrict.cpp
```diff
@@ -54,8 +69,44 @@ Coordi ToolHelperRestrict::get_coord_restrict(const Coordi &old, const Coordi &c
...
+        return find_45deg_coord(old, cur);
     }
     return cur;
 }
 
+Coordi ToolHelperRestrict::find_45deg_coord(const Coordi &old, const Coordi &cur) const
+{    
+    Coordi result;
+    Coordi vect = cur - old;
+    Coordi vect_abs = {std::abs(vect.x), std::abs(vect.y)};
```
carrotIndustries, line 82: Can be made const:
```suggestion
    const Coordi vect = cur - old;
    const Coordi vect_abs = {std::abs(vect.x), std::abs(vect.y)};
```

src/core/tools/tool_helper_restrict.cpp
```diff
@@ -54,8 +69,44 @@ Coordi ToolHelperRestrict::get_coord_restrict(const Coordi &old, const Coordi &c
         return {cur.x, old.y};
     case RestrictMode::Y:
         return {old.x, cur.y};
+    case RestrictMode::DEG45:
+        return find_45deg_coord(old, cur);
     }
     return cur;
 }
 
+Coordi ToolHelperRestrict::find_45deg_coord(const Coordi &old, const Coordi &cur) const
```
carrotIndustries, line 78: This function doesn't need to be a method of `ToolHelperRestrict`. It can be static function.

carrotIndustries:
Excellent! Works exactly as expected.

src/core/tools/tool_helper_restrict.cpp
kamtar: did you mean to say const or did you mean to make them static?

carrotIndustries: This function doesn't need access to any private members of the class, so it can be free static function:
```C++
static Coordi ToolHelperRestrict::find_45deg_coord(const Coordi &old, const Coordi &cur)
```

src/core/tools/tool_helper_restrict.hpp
```diff
@@ -12,6 +12,6 @@ class ToolHelperRestrict {
     enum class RestrictMode { X, Y, ARB, DEG45 };
     RestrictMode restrict_mode = RestrictMode::ARB;
     Coordi get_coord_restrict(const Coordi &old, const Coordi &cur) const;
-    Coordi find_45deg_coord(const Coordi &old, const Coordi &cur) const;
+    static Coordi find_45deg_coord(const Coordi &old, const Coordi &cur);
```
carrotIndustries, line 15: doesn't have to be in the class, can be a free function in the `.cpp` file

PR #628, opened 2021-11-28
Add Airwire rules

guserav:
These rules allow the user to specify components/parts that should be
considered as shortened for airwires. For example a 0 Ohm bridge
required for 1 layer boards should be considered as a connection and
result in an airwire spanning across it.


guserav:
See #629

Issue #630, opened 2021-11-30
Silk screen not rendered over tented via in 3d view


frmdstryr:
If silkscreen is placed over a tented via the center is omitted in the 3d view.

![image](https://user-images.githubusercontent.com/380158/144074965-21732789-3801-4ad4-af29-ee88a8812346.png)

carrotIndustries:
Right now, holes are subtracted from all layers except solder mask. What'd be the correct way to achieve your desired effect? Intersect solder mask (it's negative) with holes before subtracting them from silkscreen?

frmdstryr:
I was hoping to contribute a PR but after looking at the code I was not sure how to determine if the hole is for a tented via or not as that seems to come from the pool? 

The easiest approach (if my understanding is correct...) would be to add some way to skip adding the hole to the subtraction path at https://github.com/horizon-eda/horizon/blob/master/src/canvas3d/canvas_mesh.cpp#L188.  I haven't spent enough time in the code to know the best approach.  

Another case to consider that would need a similar check would be blind and buried vias.

Perhaps the hole could have a stop and start layer parameters? So the start layer of a hole could be less than the top silkscreen for a tented via.

carrotIndustries:
> Another case to consider that would need a similar check would be blind and buried vias.

These aren't supported at all right now.

We shouldn't look at the via it self, only at the resulting geometry. Let's think of it that way: Silkscreen doesn't float in air (apart from when it extends beyond the board's outline), so there must be something underneath it. 

| Hole | Solder mask (inverted, there is no soldermask where the soldermask layer is) | Silkscreen |
|--|--|--|
|Yes|Yes| Hole not covered by soldermask, so no |
|Yes|No| Hole covered by soldermask, so yes|
|No|Yes| Not a hole, so yes|
|No|no| Not a hole, so yes|

> I haven't spent enough time in the code to know the best approach.

For silkscreen, union all holes and intersect the union with the solder mask layer.

frmdstryr:
Somehow I missed the last part of your first comment... that makes sense.

RX14:
close this issue now the PR's been merged?

frmdstryr:
Yes, thank you!

PR #634, opened 2021-12-06
3d view: Keep silkscreen over tented vias for #630

frmdstryr:
I couldn't figure out how to invert the solder mask layer... but I think this gives the same result? 

image

Should it also subtract the paste layer?


carrotIndustries:
> I couldn't figure out how to invert the solder mask layer... but I think this gives the same result?

Looks good to me. Instead of operating on the inverted soldermask that has been generated by subtracting the soldermask from the outline, you could use the soldermask from the patches.

> Should it also subtract the paste layer?

While it would look prettier, it doesn't make much sense from a manufacturing point of view. #633 should take care of this.

frmdstryr:
I see what you mean now. Hopefully this is good to go.

PR #638, opened 2021-12-25
Add touchpad gesture rotation with configurable snap threshold

unrelentingtech:
Because why not. Certainly feels more convenient to me than going into the menu thing :)
To prevent accidental rotation while just zooming, a relatively large snap threshold is used by default.
If someone wishes to *completely* avoid the rotation, a huge threshold could be set.

(Other gestures are handled in the canvas but maybe they should be here in imp_layer too, e.g. if zoom would get a similar UI element as rotation has…)

### DIFF
--- a/src/imp/imp_layer.cpp
+++ b/src/imp/imp_layer.cpp
@@ -138,6 +138,20 @@ void ImpLayer::add_view_angle_actions()
             false);
     box->pack_start(*view_angle_button, true, true, 0);
 
+    gesture_rotate = Gtk::GestureRotate::create(*canvas);
+    gesture_rotate->signal_begin().connect([this](GdkEventSequence *seq) { gesture_rotate_angle_orig = view_angle; });
+    gesture_rotate->signal_update().connect([this](GdkEventSequence *seq) {
+        auto angle = gesture_rotate_angle_orig - angle_from_rad(gesture_rotate->get_angle_delta());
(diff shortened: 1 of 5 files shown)


carrotIndustries:
What a coincidence, a Gtk just gained support for touchpad gestures some days ago.

 - This gesture also applies to touchscreens, so this should be reflected in the UI and someone should test it.
 - Rotating clockwise starting from 0° doesn't snap, probably some angle wrapping thing
 - I'm not sure if we should expose the threshold setting, I'd much prefer an on/off toggle

Apart from that, it works great. 

Anyhow, my plan is to release version 2.2 in January, so I'll probably merge this once the release is out the door so that this feature can see some more testing before it makes it into a release.

unrelentingtech:
> clockwise starting from 0° doesn't snap

Oh — I could notice some "doesn't snap" situation but didn't realize what it was. Yeah, I guess I need to do the same wrapping  `set_view_angle` does here.

> someone should test it

I technically have a touchscreen laptop… except due to some hardware issue it started to not work most of the time (sometimes I can get a very lucky boot where it does work). I could make my touchpad pretend to be a touchscreen for testing, I guess :D

carrotIndustries:
*bump* I think that this is a useful thing to have.

unrelentingtech:
Yeah. I have noticed though that it does sometimes rotate when I just want to zoom. Maybe I need to add a locking mechanism that makes it so that if there was no significant rotation after the first N events, rotation is disabled this time.

PR #639, opened 2021-12-25
Tools: select connected lines, merge duplicate junctions, lines to tracks (for importing copper from DXF)

unrelentingtech:
Why? Well, say you want to use [this PCB coil generator by Texas Instruments](https://webench.ti.com/wb5/LDC/#/spirals). It exports to a few proprietary formats, and KiCad can open EAGLE projects. So you load that in KiCad, but now you need to get that into Horizon. Horizon only imports DXF, and KiCad exports a bunch of weird stuff and SVG, so you export SVG, do a slight cleanup in Inkscape (ungroup, merge all segments into one SVG path), save as DXF R12, and finally import (and scale down by a factor of like 10000 or something because formats).

But it's just lines, and somewhere along the process they became not entirely connected even. >_<

Now, just a few clicks to make it good:

https://user-images.githubusercontent.com/208340/147393195-8216b1ff-9b43-41ec-a971-a5aa2d598236.mp4

(Yeah, this particular use case could be obsoleted by KiCad PCB import or something, but there could be other fun ways to use this. Maybe someone out there wants to experiment with drawing tracks that look like fancy shapes in Inkscape or whatnot :D)

---

Naming is a hard problem :/ It's not just "one line to track", but it's not a "line loop" (though there is support for loops inside the selection!), it's a line… just… collection? sequence? something?. Also, "duplicate" is not the best word for "junctions with almost exactly the same position" I guess.

---

The merge currently does not properly support junctions attached to special board/schematic items. We should probably use like virtual methods or something so that the junction subclasses could "replace self in all my connected thingies" no matter what subclass-introduced thingies there are, if that makes sense.

Also, TODO: adjustable threshold for merge?

---

P.S. my first time doing something in the actually EDA-related code, I really like the codebase! \o/


src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <map>
+#include "document/idocument.hpp"
+
+namespace horizon {
+
+bool ToolMergeDuplicateJunctions::can_begin()
+{
+    if (!(doc.r->has_object_type(ObjectType::JUNCTION)))
+        return false;
+    for (const auto &it : selection) {
```
carrotIndustries, line 20: You can use `sel_count_type` from `selection_util.hpp`

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,89 @@
...
+        if (it.type == ObjectType::JUNCTION) {
+            return true;
+        }
+    }
+    return false;
+}
+
+ToolResponse ToolMergeDuplicateJunctions::begin(const ToolArgs &args)
+{
+    std::cout << "tool merge duplicate junctions\n";
```
carrotIndustries, line 30: we don't need that anymore

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    for (const auto &it : selection) {
+        if (it.type != ObjectType::JUNCTION) {
+            continue;
+        }
+        auto *it_j = doc.r->get_junction(it.uuid);
+        Coordd rounded(std::floor(it_j->position.x / 1e4) * 1e4, std::floor(it_j->position.y / 1e4) * 1e4);
+        if (juncs.count(rounded) == 0) {
+            juncs[rounded] = {};
+        }
```
carrotIndustries, line 42: `operator[]` does autovivification so you don't need these lines

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,89 @@
...
+                if (arc->center == junc) {
+                    arc->center = result_ptr;
+                }
+                if (arc->to == junc) {
+                    arc->to = result_ptr;
+                }
+            }
+            doc.r->delete_junction(junc_uuid);
+        }
+        // TODO: proper BoardJunction/SchematicJunction support (more than lines & arcs)
```
carrotIndustries, line 79: do we make sure that his is doesn't break if the user selected a junction to a track?

src/core/tools/tool_line_to_track.cpp
```diff
@@ -0,0 +1,141 @@
...
+bool ToolLineToTrack::can_begin()
+{
+    if (!(doc.r->has_object_type(ObjectType::LINE) && doc.r->has_object_type(ObjectType::JUNCTION)))
+        return false;
+    for (const auto &it : selection) {
+        if (it.type == ObjectType::LINE) {
+            return true;
+        }
+    }
+    return false;
```
carrotIndustries, line 24: use `sel_count_type`

carrotIndustries:
I'd have the line-to-track tool to just convert selected lines to tracks and that's it. Why does it do more?

src/core/tools/tool_line_to_track.cpp
unrelentingtech: I saw it done this way somewhere else, and I think it's for efficiency — count would keep iterating through all of them, while this returns as soon as it sees one. Maybe there should be a `sel_contains_type` helper for this?

src/core/tools/tool_merge_duplicate_junctions.cpp
unrelentingtech: It should break currently, and should work if the TODO is implemented. Maybe for now we could just ensure that there are none of those in the selection?

unrelentingtech:
> I'd have the line-to-track tool expected to just convert selected lines to tracks and that's it. Why does it do more?

Basically because I could not figure out a way to connect the resulting net-less tracks to a net :D Just figured out that "draw track" *does* work for this, but that's really not obvious.

Also at some point the tracks didn't even show up (or failed in some other way, I don't remember) without a track connecting to a net, but I think that was only when I *did* set a net but didn't connect to it.

Now it doesn't do that anymore, I agree that it shouldn't do more. But I'm kinda sad to see my fancy connect-closest-line-to-target code go lol.

src/core/tools/tool_line_to_track.cpp
unrelentingtech: @carrotIndustries ping, final decision on this little thing? :)

carrotIndustries: A helper function would be the best option.

src/core/tools/tool_line_to_track.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <functional>
+#include <vector>
+#include <unordered_set>
+#include "document/idocument_board.hpp"
+
+namespace horizon {
+
+bool ToolLineToTrack::can_begin()
+{
+    if (!(doc.r->has_object_type(ObjectType::LINE) && doc.r->has_object_type(ObjectType::JUNCTION)))
```
carrotIndustries, line 17: Since this tool only makes sense on boards, this could be simplified to

```suggestion
    if (!doc.b)
```

src/core/tools/tool_line_to_track.cpp
```diff
@@ -0,0 +1,72 @@
...
+        auto *ln = doc.r->get_line(to_visit.back());
+        to_visit.pop_back();
+        visited.insert(ln->uuid);
+
+        auto uu = UUID::random();
+        auto *track = &brd.tracks.emplace(uu, uu).first->second;
+        auto *from_junc = &brd.junctions.at(ln->from.uuid);
+        auto *to_junc = &brd.junctions.at(ln->to.uuid);
+        if (!from_junc->only_lines_arcs_connected() || !to_junc->only_lines_arcs_connected()) {
+            imp->tool_bar_flash(
```
carrotIndustries, line 50: Why do we need this limitation?

src/core/tools/tool_line_to_track.cpp
```diff
@@ -0,0 +1,72 @@
...
+ToolResponse ToolLineToTrack::begin(const ToolArgs &args)
+{
+    auto &brd = *doc.b->get_board();
+
+    auto sel_line_uuid = sel_find_one(selection, ObjectType::LINE).uuid;
+
+    std::unordered_set<UUID> visited;
+    std::vector<UUID> to_visit = {sel_line_uuid};
+
+    while (!to_visit.empty()) {
```
carrotIndustries, line 36: I think it makes sense to split this tool into two tools:
 - Select connected lines
 - A tool that just converts selected lines to tracks

src/core/tools/tool_line_to_track.cpp
unrelentingtech: I'm not sure what would happen when lines are connected to board elements. hm, probably nothing too bad actually?

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,88 @@
...
+{
+    std::map<Coordd, std::vector<UUID>> juncs;
+
+    for (const auto &it : selection) {
+        if (it.type != ObjectType::JUNCTION) {
+            continue;
+        }
+        auto *junc = doc.r->get_junction(it.uuid);
+        if (!junc->only_lines_arcs_connected()) {
+            imp->tool_bar_flash("cannot merge junctions that are connected to board elements (not just lines/arcs)");
```
carrotIndustries, line 38: since this tool also applies to schematics, it shouldn't mention board items

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,88 @@
...
+        if (it.type == ObjectType::JUNCTION) {
+            return true;
+        }
+    }
+    return false;
+}
+
+ToolResponse ToolMergeDuplicateJunctions::begin(const ToolArgs &args)
+{
+    std::map<Coordd, std::vector<UUID>> juncs;
```
carrotIndustries, line 30: why not use `Coordi` as the key? should avoid rounding errors

unrelentingtech: How would the key calculation look with integers? I guess just `/ 1e4 * 1e4` would accomplish this since the integer division would basically do a `floor` type thing, right?

carrotIndustries: since you don't need the coordinates for anything else, you might as well skip the multiplication afterwards and just to `/0.01_mm` if you want to be extra-accurate you could add .005mm beforehand to get round behaviour rather than truncation

unrelentingtech:
All split now!

src/core/tools/tool_select_connected_lines.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "core/tool.hpp"
+
+namespace horizon {
+
+class ToolSelectConnectedLines : public ToolBase {
+public:
+    using ToolBase::ToolBase;
+    ToolResponse begin(const ToolArgs &args) override;
+    ToolResponse update(const ToolArgs &args) override;
+    bool can_begin() override;
```
carrotIndustries, line 11: needs `is_specific` to return true so that it shows up in the context menu

src/core/tools/tool_select_connected_lines.cpp
```diff
@@ -0,0 +1,54 @@
...
+        auto *to_junc = doc.r->get_junction(ln->to.uuid);
+
+        std::copy(from_junc->connected_lines.begin(), from_junc->connected_lines.end(), std::back_inserter(to_visit));
+        std::copy(to_junc->connected_lines.begin(), to_junc->connected_lines.end(), std::back_inserter(to_visit));
+    }
+
+    if (visited.size() == 1)
+        imp->tool_bar_flash("no connected lines found");
+
+    return ToolResponse::commit();
```
carrotIndustries, line 47: you can return `ToolResponse::end` since this tool didn't touch anything. this way, we don't get an undo entry

you should also call `imp->get_canvas()->set_selection_mode(CanvasGL::SelectionMode::NORMAL);` to leave hover select mode, since moving the cursor a tiny bit could unselect everything

src/core/tools/tool_merge_duplicate_junctions.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <map>
+#include "document/idocument.hpp"
+
+namespace horizon {
+
+bool ToolMergeDuplicateJunctions::can_begin()
+{
+    if (!(doc.r->has_object_type(ObjectType::JUNCTION)))
+        return false;
+    return sel_has_type(selection, ObjectType::JUNCTION);
```
carrotIndustries, line 21: merging only makes sense if we have more than one junction

src/core/tools/tool_lines_to_tracks.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "core/tool.hpp"
+
+namespace horizon {
+
+class ToolLinesToTracks : public ToolBase {
+public:
+    using ToolBase::ToolBase;
+    ToolResponse begin(const ToolArgs &args) override;
+    ToolResponse update(const ToolArgs &args) override;
+    bool can_begin() override;
```
carrotIndustries, line 11: also add is_specific

src/core/tools/tool_merge_duplicate_junctions.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "core/tool.hpp"
+
+namespace horizon {
+
+class ToolMergeDuplicateJunctions : public ToolBase {
+public:
+    using ToolBase::ToolBase;
+    ToolResponse begin(const ToolArgs &args) override;
+    ToolResponse update(const ToolArgs &args) override;
+    bool can_begin() override;
```
carrotIndustries, line 11: add is_specific

src/core/tools/tool_lines_to_tracks.cpp
```diff
@@ -0,0 +1,52 @@
...
+            // lots of potential to e.g. create tracks between different nets
+            imp->tool_bar_flash(
+                    "cannot convert a line that is already connected to board elements (not just lines/arcs)");
+            return ToolResponse::revert();
+        }
+        track->to.connect(to_junc);
+        track->from.connect(from_junc);
+        track->width_from_rules = false;
+        track->width = ln->width;
+        track->layer = args.work_layer;
```
carrotIndustries, line 41: why not `ln->layer`? you should also make sure to not create tracks on non-copper layers as these are impossible to get rid of

carrotIndustries:
Works great now. just one more thing. The merge duplicate junctions tool might be a bit engimatic as there's no real indication of what it did or if it even did anything. Use `imp->tool_bar_flash` to either indicate that the tool did nothing (and in that case return `ToolResponse::end()`) or say how many unique junctions there are now. Use `object_descriptions.at(ObjectType::JUNCTION).get_name_for_n` to get the appropriate sigular/plural form.

unrelentingtech:
Good catch!

> get_name_for_n

Ooh, nice to know. It's a bit annoying in this case that it always capitalizes "Junction" but whatever.

PR #644, opened 2022-01-04
padstack: allow set-shape position on holes

unrelentingtech:
Holes have a position too :) Maybe `set-position` should be a dedicated command actually since it's not specifically *shape* related…

### DIFF
--- a/src/pool/padstack.cpp
+++ b/src/pool/padstack.cpp
@@ -107,6 +107,16 @@ std::optional<std::string> Padstack::MyParameterProgram::set_hole(const TokenCom
             }
         }
     }
+    else if (shape == "position") {
+        int64_t x, y;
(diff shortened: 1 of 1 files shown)


carrotIndustries:
It makes more sense to add this as `set-hole [ foo position ]`. That way, it's consistent with `set-shape`

unrelentingtech:
Right, that works, done

unrelentingtech:
btw this is what I needed this for: an SMD-or-TH padstack to make a footprint like [this kicad one](https://github.com/keebio/Keebio-Parts.pretty/blob/c7ae3b44674679f4d767767c002fed1eacd414a1/Diode-Hybrid-Back.kicad_mod) with only two pads!

Screen_2022-01-04-18:12:27

PR #650, opened 2022-01-24
Optimize performance of check_clearance_copper

guserav:
Main optimisations:
It is now ensured that every polygon that needs expanding will only be
expanded once. Additionally the number of unique polygons actually
expanded should be reduced.

The later is achieved by a primitive implementation of set-cover. Which
is running a single pass over all patch_pairs and if no polygon is
already expanded by the required clearance the polygon with more
neighbours is used.

Additionally only patch_pairs that have overlapping bounding boxes will
be scheduled for parallel checking of polygon intersection. This had
the upside of also calculating the bounding boxes only once.

Other optimisations:
- Pre-filter patches and calculate patch pairs in a more structured
  approach.
- Trusting the bounding box check more, not using twice the clearance
  only clearance + 10 (Even though I don't see the point in even doing
  that, but carrotIndustries insisted...).

Bug fixes on the way:
- Copper around a via is no longer matched against copper on all other
  layers
- Intersection PTHs will be reported. This is that big a deal as PTHs
  should always be surrounded by copper which reported the error of e.g.
  overlapping vias previously.

### DIFF
--- a/src/board/board_rules_check.cpp
+++ b/src/board/board_rules_check.cpp
@@ -134,103 +134,135 @@ RulesCheckResult BoardRules::check_hole_size(const Board &brd) const
     return r;
 }
 
-static std::pair<ClipperLib::IntPoint, ClipperLib::IntPoint> get_patch_bb(const ClipperLib::Paths &patch)
+inline void get_patch_bb(const ClipperLib::Paths &patch, ClipperLib::IntRect &out)
 {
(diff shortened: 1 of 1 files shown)


carrotIndustries:
After some offline discussions, we arrived at https://github.com/horizon-eda/horizon/commit/2718b88614eb1106e8acf2314ebcad2633508848

PR #669, opened 2022-06-01
Close path for each plane patch for #668

### LINKED ISSUE #668: Copper export to pdf is missing some plane segments
When exporting the copper layers to pdf, the ground plane is missing a few segments at the bottom of pads.

image

At first I thought it was a rending thing but if you open the pdf in inkscape the lines are indeed missing. It is in both fill and outline mode but is not visible in the fill mode.

frmdstryr:
Fixes #668 
image

I'm unsure if there's any unintended side effects of this.


carrotIndustries:
> I'm unsure if there's any unintended side effects of this.

I don't see any, the existing code did the wrong thing.

PR #686, opened 2022-09-01
Phong shader for 3d view

frmdstryr:
This adds phong shading to the face and cover shaders.  

image

I'm not sure if updating the `cam_position` in base render is ok or if it should be done somewhere else? 

Also maybe it should support multiple lights since it only highlights at one angle. Also the layers could have adjustable parameters but I'm not sure where that should be added.

Old view for reference
horizon-no-shading

### DIFF
--- a/src/canvas3d/canvas3d_base.cpp
+++ b/src/canvas3d/canvas3d_base.cpp
@@ -265,7 +265,6 @@ void Canvas3DBase::render(RenderBackground mode)
     float theta = glm::radians(90 - cam_elevation);
     auto cam_offset = glm::vec3(r * sin(theta) * cos(phi), r * sin(theta) * sin(phi), r * cos(theta));
     auto cam_pos = cam_offset + glm::vec3(center, 0);
-    cam_position = cam_pos;
 
     glm::vec3 right(sin(phi - 3.14f / 2.0f), cos(phi - 3.14f / 2.0f), 0);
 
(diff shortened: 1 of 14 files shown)


carrotIndustries:
Cool, I've been thinking about this as well, but could never convince myself that it's worth the effort.

While it does indeed look good at many angles, there still are some cases where phong shading looks unexpected.
2022-09-07-225117_1012x806_scrot

Also, zooming in and out causes unintuitive behavior as it looks as if the light source is changing position.

Btw, there's no 'r' in phong.

frmdstryr:
The light now follows the camera. I made it slightly offset so it's not as plain. Hopefully that's better.

image

carrotIndustries:
I think you're headed in the right direction, but some things still need improvement before we can merge this:

 - There should be less settings for the user, i.e. an on/off toggle or a single shinyness(?) slider.
 - Brightly-colored models shouldn't turn into an amorphous white blob when looked at the wrong way:
 
2022-09-21-000017_1032x642_scrot

The goal is that the new shading is an improvement in all use cases and should not require any tweaking on the user's end to result in something that looks good.

frmdstryr:
Some of the model face normals still seem to be incorrect. I could not find anything obvious as to why.

image

frmdstryr:
Much better now :)
image

bob-u:
Shouldn't the white 3D surfaces above be of different shades of gray (silver)?

frmdstryr:
The light is behind the camera (slightly offset). If you adjust `light_theta` in canvas3d_base.cpp to `auto light_theta = theta - glm::radians(45.0);` then it is more of an overhead light. 

Maybe there should be user settings to change the light position and adjust settings? On my setup it looks slightly different based on which display the window is on. 

image

No shading
horizon-eda-standard

Shading
horizon-eda-shaded

Or maybe it's not worth the extra complexity for slightly different colors, idk.

PR #687, opened 2022-09-02
Separate multi labeled pins for better readability

oxplot:
*Initial discussion on [discourse](https://horizon-eda.discourse.group/t/separating-alternate-pin-names-with-slash/145).*

Currently, when showing multiple alternate pin names, they are separated by a single space on the schematics:

image

This is somewhat difficult to read at times and may be mistaken in some scenarios as a single label. This PR adds unicode MIDDLE DOT as a visual separator between the labels:

Screenshot from 2022-09-02 17-34-56

### DIFF
--- a/src/schematic/schematic_symbol.cpp
+++ b/src/schematic/schematic_symbol.cpp
@@ -9,7 +9,7 @@
 
 namespace horizon {
 
-const char *pin_name_sep = " · ";
+static const char *pin_name_sep = " · ";
 
 static const LutEnumStr<SchematicSymbol::PinDisplayMode> pdm_lut = {
         {"selected_only", SchematicSymbol::PinDisplayMode::SELECTED_ONLY},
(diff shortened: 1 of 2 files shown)


carrotIndustries:
That's is indeed more readable than just a single space. There are some other code paths, most notably for the "all" pin display mode, that also need to be changed to match the new style in SchematicSymbol::apply_pin_names.

carrotIndustries:
LGTM apart from the failing style check.

carrotIndustries:
Sorry, just one more thing: How about factoring the separator out into a string constant?

src/schematic/schematic_symbol.cpp
```diff
@@ -9,6 +9,8 @@
 
 namespace horizon {
 
+const char *pin_name_sep = " · ";
```
carrotIndustries, line 12:
```suggestion
static const char *pin_name_sep = " · ";
```

PR #688, opened 2022-09-07
Initial local db stock provider

frmdstryr:
This adds a stock info provider using a local sqlite database. It is currently set so right clicking the stock column in the part browser opens a dialog to edit the stock information.  I'm not sure how to get the part MRN for the dialog.

Any feedback would be appreciated. 

image

### DIFF
--- a/src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
+++ b/src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
@@ -211,7 +211,7 @@ void PoolGitBox::status_cb(const char *path, unsigned int status_flags)
     if (!(status_flags & GIT_STATUS_IGNORED)) {
         q_status->reset();
         q_status->bind(1, std::string(path));
-        q_status->bind(2, status_flags);
+        q_status->bind(2, (int)status_flags);
         q_status->step();
     }
 }
--- a/src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
+++ b/src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
@@ -211,7 +211,7 @@ void PoolGitBox::status_cb(const char *path, unsigned int status_flags)
     if (!(status_flags & GIT_STATUS_IGNORED)) {
         q_status->reset();
         q_status->bind(1, std::string(path));
-        q_status->bind(2, (int)status_flags);
+        q_status->bind(2, status_flags);
         q_status->step();
     }
 }
--- a/src/util/stock_info_provider_local.cpp
+++ b/src/util/stock_info_provider_local.cpp
@@ -1,6 +1,5 @@
 #include "stock_info_provider_local.hpp"
 #include "stock_info_provider_local_editor.hpp"
-#include <thread>
 #include "nlohmann/json.hpp"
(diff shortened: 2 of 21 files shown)


src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
```diff
@@ -211,7 +211,7 @@ void PoolGitBox::status_cb(const char *path, unsigned int status_flags)
     if (!(status_flags & GIT_STATUS_IGNORED)) {
         q_status->reset();
         q_status->bind(1, std::string(path));
-        q_status->bind(2, status_flags);
+        q_status->bind(2, (int)status_flags);
```
carrotIndustries, line 214: why this change?

src/util/stock_info_provider_local.cpp
```diff
@@ -0,0 +1,403 @@
...
+            auto data = (const char *)bytes->get_data(size);
+            stock_db.execute(data);
+        }
+    }
+    LocalStockPreferences prefs;
+    SQLite::Database stock_db;
+};
+
+
+class StockInfoProviderLocalWorker {
```
carrotIndustries, line 80: I think we can don't need that complexity for the local stock info provider as a local database is fast enough to be queried from the UI thread.

src/util/stock_info_provider_local.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+class LocalStockManager {
+public:
+    static LocalStockManager &get()
+    {
+        static LocalStockManager inst;
+        return inst;
+    }
+
+    std::pair<bool, json> query(const UUID &part)
```
carrotIndustries, line 28: why not return a `StockInfoLocalRecord` here?

frmdstryr:
Well it still seems slower than the worker thread version... the only other thing I can think of trying yet is to simply ignore the parts list, do a select all and filter it manually unless you have any other ideas.

RX14:
I think this PR is great, but has there been thought for laying groundwork to add plugins in the future for other methods of keeping stock? I have an existing local inventory system and It'd be nice to integrate that in the future.

frmdstryr:
Could you have a script/cron job or something update/generate the local database?

carrotIndustries:
> I think this PR is great, but has there been thought for laying groundwork to add plugins in the future for other methods of keeping stock? I have an existing local inventory system and It'd be nice to integrate that in the future.

I see several ways we could do this:

 - define a C API and use plain old dlopen/dlsym to load the plugin
 - use libpeas (the only thing I know about is is the name and it should make plugins easier)
 - something based on HTTP/REST
 - something based on python

Any of this option will result in something that'll work in some way, we just have to be aware of the tradeoffs.

RX14:
I don't want to delay this PR to deal with this issue, AFAIK there's nothing here that prevents implementing a plugin API of some kind in the future. My personal preference would be some kind of interpreter-based plugins, but C plugins would also allow that to be added.

carrotIndustries:
I'm planning to do the 2.4 release within the next month or so and I think we already have enough features for it and I don't want to risk introducing bugs with new features. I'll have another look at this PR once the release is done.

PR #703, opened 2023-01-17
Show pin count in unit editor

frmdstryr:
image


src/pool-prj-mgr/pool-mgr/editors/unit_editor.hpp
```diff
@@ -35,6 +35,7 @@ class UnitEditor : public Gtk::Box, public PoolEditorBase {
     Gtk::Button *add_button = nullptr;
     Gtk::Button *delete_button = nullptr;
     Gtk::CheckButton *cross_probing_cb = nullptr;
+    Gtk::Label *pin_count = nullptr;
```
carrotIndustries, line 38: should be `pin_count_label`

carrotIndustries:
LGTM apart from the one review comment, use `label_set_tnum` to make the label use tabular figures so that the numbers don't move when they shouldn't.

PR #714, opened 2023-04-02
Test dll build issue

fk0815:
Testing only


fk0815:
@carrotIndustries it seems that there is an issue with the pll check script. For a test I did just a change in a documentation file, the ckeck fails anyway. Could you have a look on that? I do not have a Windows build ready to use.

carrotIndustries:
Yeah, this happens from time to time since the CI pipeline always uses the latest packages an doesn't pin versions. I've created a branch to test the fix: https://github.com/horizon-eda/horizon/compare/master...win32-dll-3

PR #717, opened 2023-04-04
Fix build issue with gcc13

fk0815:
This fixes the "error: 'uint64_t' does not name a type" build error when using gcc13.

### DIFF
--- a/src/board/board_rules_check_util.hpp
+++ b/src/board/board_rules_check_util.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <string>
 #include "clipper/clipper.hpp"
+#include <cstdint>
 
 namespace horizon {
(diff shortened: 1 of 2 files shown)


carrotIndustries:
Which distro(?) did you test this on? I think we should add it to CI until GCC 13 is released and arch linux has caught up.

fk0815:
I found it on opensuse tumbleweed where I do a daily build on changes here: https://build.opensuse.org/package/show/home:frank_kunz/horizon-next

Let me know if you want to wait for arch linux, when I will add the patch to that build until it makes into the master branch.

carrotIndustries:
Thanks, I added opensuse tumbleweed to the CI pipeline so things like these are less likely to happen in the future.

PR #720, opened 2023-05-13
imp board: minimum diameter for holes in STEP export

eqvinox:
Loading STEP files into some other CAD software can take quite some time depending on the complexity of the STEP file.  A board with a lot of vias takes several minutes to load into FreeCAD.  While 3D component models can be excluded (worst case even by deleting individual ones after export, like all bypass caps), editing the board itself is not that easy since it is one single 3D object.

This adds a STEP export setting to exclude all holes/vias below a given diameter.  (Generally speaking, while larger holes can be important to have in the STEP export, there's generally some cut off below which you don't care, like 1mm.)

---

I had hacked this together quite some time ago, but I just hardcoded a 0.6mm limit. Took some time today to implement proper UI for it :smile:

### DIFF
--- a/src/board/step_export_settings.cpp
+++ b/src/board/step_export_settings.cpp
@@ -6,7 +6,7 @@ namespace horizon {
 
 STEPExportSettings::STEPExportSettings(const json &j)
     : filename(j.at("filename").get<std::string>()), prefix(j.at("prefix").get<std::string>()),
-      include_3d_models(j.at("include_3d_models"))
+      include_3d_models(j.at("include_3d_models")), min_diameter(j.value("min_diameter", 0_mm))
 {
 }
 
@@ -16,6 +16,8 @@ json STEPExportSettings::serialize() const
     j["filename"] = filename;
     j["prefix"] = prefix;
     j["include_3d_models"] = include_3d_models;
+    if (min_diameter)
+        j["min_diameter"] = min_diameter;
     return j;
 }
 
--- a/src/board/step_export_settings.hpp
+++ b/src/board/step_export_settings.hpp
@@ -18,5 +18,6 @@ public:
     std::string filename;
     std::string prefix;
     bool include_3d_models = true;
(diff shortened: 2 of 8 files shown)


eqvinox:
BTW: I wasn't sure about slots, it certainly would be easy to apply the limit to the long axis of a slot. But since slots are somewhat rare/"special", I thought it might be better to just keep all of them unconditionally.

carrotIndustries:
LGTM, I wouldn't have implemented it any different :+1: 

I'll merge it once version 2.5 is released.

> While 3D component models can be excluded

Another user also once requested an option to exclude certain parts from STEP export to reduce file size. I couldn't quite decide how we want to implement this:

 1. In the STEP export window, select a list of packages or models to exclude
 2. In the STEP export window, select a list of components to exclude. Maybe make that list sortable by package to get 1.
 3. On each package, add a property "Omit in STEP export"

I think option 1 should be sufficient for most peoples use case of excluding insignificant parts such as small passives.

eqvinox:
> Another user also once requested an option to exclude certain parts from STEP export to reduce file size.

Yeah… my current workflow is export from Horizon, open STEP file in FreeCAD on its own, delete all capacitors and resistors, save, then import into my "main" FreeCAD project. A bit clunky, but works…

[…]
>     1. In the STEP export window, select a list of packages or models to exclude
[…]
> I think option 1 should be sufficient for most peoples use case of excluding insignificant parts such as small passives.

ACK. Selecting by package probably has the lowest weirdness / complexity / unexpected side effects. I'll put it on my list of things to do when I hit some unexpected free time :smile:

For completeness I do want to throw in:

4. specify a minimum board area in mm², exclude anything below that (calculating it per package on either courtyard or package polygons.)

But I think that's pushing it a bit too far into "simplification" territory and might do weird things in some cases (e.g. small but tall components? Mechanical components that have no courtyard/package polygons?) Plus calculating the area from polygons is a whole bunch of code…

PR #721, opened 2023-05-17
imp package: add bottom-up STEP projection

eqvinox:
When drawing packages, while the top-down STEP projection is useful, it can in some cases be just entirely silly...  for example QFN parts in a top down view are just the package outline.

Add a bottom-up projection where you actually see useful things.

Note that this is not a rotation, it's a reversal of the Z axis.  This doesn't seem to be possible in gp_Ax2 since it turns around the handedness of the coordinate system.  Hence some transform shenanigans on this.

Bonus points for gp_Trsf trying to be clever and converting the (1,1,-1) matrix into a "(-1,-1,1) + scale -1" matrix, the scale part of which is then later ignored by default in HLRAlgo_Projector...  so suddenly it's an XY flip :'(

---

I feel like OpenCASCADE made this much harder than necessary :cry: … took a bunch of whacking things with a hammer to get this going.

For reference here's top-down and bottom-up for a USB connector, with the STEP model moved to the side to make it more obvious:

Top down (only option before this PR)

top down

Bottom up (added by this PR)

bottom up

### DIFF
--- a/src/imp/imp_package.hpp
+++ b/src/imp/imp_package.hpp
@@ -12,6 +12,14 @@ namespace STEPImporter {
 class STEPImporter;
 }
 
+/* for projecting STEP models into package editor */
+enum class ProjectionMode {
+    /* standard top-down view / projection */
+    TOP_DOWN,
+    /* Z-mirrored / inverted bottom-up projection (NB: not a rotation) */
+    BOTTOM_UP,
+};
+
 class ImpPackage : public ImpLayer {
     class ImportCanvas3D;
     friend class ModelEditor;
@@ -67,7 +75,7 @@ private:
 
         ~ModelInfo();
     };
-    void project_model(const Package::Model &model);
+    void project_model(const Package::Model &model, ProjectionMode proj);
 
     std::map<std::string, ModelInfo> model_info;
     std::mutex model_info_mutex;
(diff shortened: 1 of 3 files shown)


eqvinox:
Here's a QFN for comparison (again 3D model moved to make it easier to see), where the top-down view is in fact just a useless rectangle (and a company logo :laughing:)

Note that the pin 1 marker is in the right place — because the 3D model is *not* rotated, it's mirrored in Z axis / projection direction reversed / however you want to call it.

horizon_bottom_up_qfn

src/imp/imp_package.hpp
```diff
@@ -67,7 +67,7 @@ class ImpPackage : public ImpLayer {
 
         ~ModelInfo();
     };
-    void project_model(const Package::Model &model);
+    void project_model(const Package::Model &model, bool top_view);
```
carrotIndustries, line 70: Nitpicking: These days, I prefer to use enums over booleans in function arguments since it makes things clearer at the call sites.
I think an enum likes this makes sense:
`enum class ProjectionMode {TOP_DOWN, BOTTOM_UP};`

eqvinox: sure, gimme a few minutes to edit this over

eqvinox: done

PR #731, opened 2023-08-28
Check for unused junctions in symbol editor 

### LINKED ISSUE #730: Feature request: Add rule in symbol editor to check for junctions with no connections
I created a symbol by drawing a line rectangle and adding all the pins. The rectangle was too big so I deleted the lines and made a new line rectangle. When placed in the schematic there was a lot of padding around the symbol.

image

After going back into the symbol editor it looked fine at first. Then I selected everything and noticed there were random junctions left over that I didn't delete before. It'd be nice if the rules could check for this (and/or just auto delete them)

image

frmdstryr:
Hopefully that is all that's needed. Seems to work.

horizon-junction-connections-check

Fixes #730

### DIFF
--- a/src/symbol/symbol_rules_check.cpp
+++ b/src/symbol/symbol_rules_check.cpp
@@ -211,6 +211,19 @@ RulesCheckResult SymbolRules::check_symbol(const Symbol &sym) const
         }
     }
 
+    {
+        // Check unused junctions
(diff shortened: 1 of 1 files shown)


frmdstryr:
Hey I found a few that fail the check, so I think it works :) The "Single Buffer with enable" and single gate mosfet driver. 

image


Is there a way to have the bot run checks on everything?

Issue #657, opened 2022-05-06
Feature request: modify arc radius, possibly back to sharp vertex


tkornack:
You have a nice tool to round off vertex. It would be great to extend your round-off tool to modify an existing arc. It's great if this can make it go back to a sharp vertex, too.

PR #735, opened 2023-09-06
Update round off vertex to support modification/removal 

frmdstryr:
This updates the round off vertex tool so the radius can be modified or removed by selecting the center point and re-opening the tool. If the radius is very small (eg set to 0 by editing it) it will undo the round off.

For #657.

I originally had the revert_arc function checking that the magnitude of the radii to each point are roughly the same but it seems they can be off due to rounding so that was removed. Let me know if you find any issues.


tkornack:
This sounds great! Haven't tested yet, but I'm curious how hard is it to make it work when selecting the arc segment instead of the arc center?

src/common/common.hpp
```diff
@@ -268,6 +269,13 @@ template <typename T> class Coord {
     {
         return {x, y};
     }
+
+    std::string str() const
```
carrotIndustries, line 273: we already have `coord_to_string` in `geom_util.hpp`

carrotIndustries:
Sorry for not getting back to this, but you can keep the PR open.

PR #740, opened 2023-09-13
build: find podofo header on voidlinux

ii8:
On my system `podofo.h` is in `/usr/include/podofo/podofo.h` so it's not found by the current meson script. I've changed it to work for me, I hope it still works on systems with `podofo-0.9` directories, can you test that? I also added `/usr/local` variants for completeness.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,6 +4,7 @@ project('horizon-eda', ['cpp', 'c'],
   default_options: ['cpp_std=c++17', 'warning_level=1'],
 )
 
+fs = import('fs')
 cxx = meson.get_compiler('cpp')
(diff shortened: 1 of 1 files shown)


ii8:
From build-win32:

> meson.build:78:17: ERROR: Search directory /usr/lib/podofo-0.9 is not an absolute path.

I'm a little stumped by this, this path wasn't changed, it looks absolute to me, and I don't have a windows machine to try things

It looks like the other build failures are not related to my changes.

carrotIndustries:
Is that still a problem with the support for podofo 0.10?

ii8:
It does find the podofo 0.10 files without problem, nice ty.

PR #741, opened 2023-09-13
Fix crash when pns is walking around arcs

moritz-meier:
When routing a track in walkaround mode, horizon crashes when the routed track collides with an arc. This due to the assert statement:

`horizon-imp: ../3rd_party/router/kimath/src/geometry/shape_collisions.cpp:504: bool Collide(const SHAPE_ARC&, const SHAPE_CIRCLE&, int, int*, VECTOR2I*, VECTOR2I*): Assertion (!aMTV) failed.`

In KiCad this assert statement was removed to fix this problem, so i think it should be ok to do the same here.

https://gitlab.com/kicad/code/kicad/-/commit/df9cf0a0c39ed99527b0c04e3892e8dd7ed603e7

Would be nicer to port the entire KiCad Router 7.0.7 or at least 6.0.11 to horizon, but i tried and then decided it is to much trouble for just a small fix.

### DIFF
--- a/3rd_party/router/kimath/include/geometry/shape.h
+++ b/3rd_party/router/kimath/include/geometry/shape.h
@@ -96,6 +96,11 @@ public:
         return m_type;
     }
 
+    wxString TypeName() const
+    {
+        return SHAPE_TYPE_asString( m_type );
+    }
+
     virtual bool HasIndexableSubshapes() const
     {
         return false;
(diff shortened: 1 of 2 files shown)


carrotIndustries:
Would be nice to have this as two separate commits:

 - Remove the offending assertion
 - Fix the assertion texts

PR #746, opened 2023-09-27
odb export: include mounting type atrribute of components

ii8:
Some manufacturers (in particular Aisler) require component mounting information in the ODB++ output to do assembly.

### DIFF
--- a/src/export_odb/eda_data.cpp
+++ b/src/export_odb/eda_data.cpp
@@ -275,6 +275,7 @@ EDAData::Pin &EDAData::Package::add_pin(const horizon::Pad &pad)
         pin.type = Pin::Type::THROUGH_HOLE;
         pin.mtype = Pin::MountType::THROUGH_HOLE;
         pin.etype = Pin::ElectricalType::ELECTRICAL;
+        mtype = attribute::comp_mount_type::THMT;
         break;
 
     case Padstack::Type::TOP:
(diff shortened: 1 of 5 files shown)


ii8:
Is there something I need to do so this can be merged?

carrotIndustries:
How does this show up in the valor ODB++ viewer?

ii8:
I went here: https://trials.sw.siemens.com/en-US/trials/valor-online-pcb-viewer
But it looks like I have to wait to use it (got no email yet):
image
Bit strange, not sure if that's what you meant.

This is how it appears on aisler:
image

And without this change:
image

ii8:
I've still not heard from them, so I probably wont. Is there some other way to verify this PR? It produces output conforming to the specification here: https://odbplusplus.com//wp-content/uploads/sites/2/2021/02/odb_spec_user.pdf

src/export_odb/eda_data.cpp
```diff
@@ -275,19 +275,23 @@ EDAData::Pin &EDAData::Package::add_pin(const horizon::Pad &pad)
...
         pin.etype = Pin::ElectricalType::ELECTRICAL;
+        mtype = attribute::comp_mount_type::THMT;
         break;
 
     case Padstack::Type::TOP:
     case Padstack::Type::BOTTOM:
         pin.type = Pin::Type::SURFACE;
         pin.mtype = Pin::MountType::SMT;
         pin.etype = Pin::ElectricalType::ELECTRICAL;
+        if (!mtype)
```
carrotIndustries, line 286: So this there to coalesce SMT+THMT to THMT?

ii8: Yep

carrotIndustries:
Thanks for the nudge. It really shows that the I wrote the ODB++ exporter in one go while trying to be smart with templates and such and have largely forgotten how it's put together ;( Had a look at the code and don't see anything wrong with it. ODB++ diff also looks as expected.

PR #747, opened 2023-10-01
add support for OpenBSD

dl6tom:
since OpenBSD has no API to get the path that was used to create the current process, the get_exe_dir() implementation makes use of argv[0] and the PATH environment variable.

### DIFF
--- a/src/pool-prj-mgr/pool-prj-mgr-main.cpp
+++ b/src/pool-prj-mgr/pool-prj-mgr-main.cpp
@@ -11,6 +11,11 @@
 
 int main(int argc, char *argv[])
 {
+#ifdef __OpenBSD__
+    extern char *argv0;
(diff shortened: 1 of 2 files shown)


carrotIndustries:
Is there any CI service that supports OpenBSD so we can make sure that this code at least compiles?

dl6tom:
i have it running here, so can assure you it build and runs. it seems sourcehut supports OpenBSD in their CI (https://man.sr.ht/builds.sr.ht/compatibility.md). i can take a look what it takes to set this up.

carrotIndustries:
I could set up a webhook for that purpose.

dl6tom:
i think that would be https://jenkins.0x90.space/github-webhook/ then. not sure how to report the build status back to github.

PR #755, opened 2023-10-28
Add support for libpodofo version >= 0.10.0

fk0815:
This adds support for new libpodofo 0.10.0 or newer and keeps legacy support as well.

Build here: https://build.opensuse.org/package/show/home:frank_kunz/horizon-next

tested on openSUSE Tumbleweed (new libpodofo) and 15.4/15.5 (old libpodofo).


carrotIndustries:
Thanks a lot for taking this on!

I tried it on my system and it crashed since `document.GetFonts().SearchFont("Helvetica")` returns a `nullptr`.

I then replaced it with `GetStandard14Font(PoDoFo::PdfStandard14FontType::Helvetica)`. This then didn't segfault but throws an exception when trying to draw non-ascii(?) characters:

````
PdfErrorCode::InvalidFontData, The font data is invalid.
Callstack:t#0 Error Source: main/PdfEncoding.cpp(82), Information: The provided string can't be converted to CID encoding
````

fk0815:
I tried that as well, but I did not get any error with your proposed change. I used some umlauts (ÄÖÜ) added as text to the schematic and that works for me. How do you add non ascii chars? What is the operating system?

carrotIndustries:
> How do you add non ascii chars? What is the operating system?

In my case it was the ohm symbol from a resistor value. OS is Arch Linux.

carrotIndustries:
Arch Linux having dropped legacy podofo motivated me to revisit this PR. I fixed the font issue by bundling it.

Thanks a lot for your contributions! :heart:

fk0815:
Is there a new release in the queue? Or would it make sense to backport that to 2.5.0 ?

carrotIndustries:
Good question, it's almost been a year since the last release, so it's about time for a new one and there are new features worth it.

PR #766, opened 2024-03-09
Add package and button to reset placed list to place list window

frmdstryr:
Screenshot_20240309_112934


src/imp/parts_window.cpp
```diff
@@ -44,6 +49,7 @@ PartsWindow::PartsWindow(const Board &brd) : Gtk::Window(), board(brd), state_st
     tree_view->get_column(1)->set_sort_column(list_columns.qty);
     tree_view->append_column("Value", list_columns.value);
     tree_view->append_column("MPN", list_columns.MPN);
+    tree_view->append_column("Package", list_columns.package);
```
carrotIndustries, line 52: maybe use `tree_view_append_column_ellipsis`  so that long package names get ellipsized.

Issue #764, opened 2024-02-26
Horizon does not support OpenCascade 7.8.0


thierry-FreeBSD:
The build fails with the message:
```
===>   horizon-eda-2.5.0_3 depends on shared library: libTKSTEP.so - not found
*** Error code 1
```
There is no more libTKSTEP in OpenCascade 7.8.0.

bob-u:
From release notes (https://dev.opencascade.org/doc/overview/html/occt__upgrade.html):

Reorganized DE TK

DE TK components have been combined or separated based on specific CAD formats to support plug-in ability.

    Components now have a "TKDE" prefix. The available list includes TKDESTEP, TKDEOBJ, TKDEIGES, TKDEGLTF, TKDEVRML, TKDEPLY, TKDESTL.
    The DE DRAW TK has been updated in a similar way: DRAW components now have a "TKXSDRAW" prefix. The available list includes TKXSDRAWSTEP, TKXSDRAWOBJ, TKXSDRAWIGES, TKXSDRAWGLTF, TKXSDRAWVRML, TKXSDRAWPLY, TKXSDRAWSTL.

Migration problems may occur during configuration time or compile time. Ensure that you update your project configuration accordingly.

thierry-FreeBSD:
> None of the distros we have CI for already has opencascade 7.8 Any ideas how to reproduce this?

I guess that most distros won't upgrade OpenCascade as long as it breaks its consumers - at least this is what we are doing in FreeBSD: see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277294 .

Thus, ATM, you have to compile it yourself.

frmdstryr:
I'm using a locally built horizon with locally built OCCT 7.8.0 and it works fine here... didn't have to change anything.

```
$ lddtree .local/bin/horizon-imp
horizon-imp => .local/bin/horizon-imp (interpreter => /lib64/ld-linux-x86-64.so.2)
  ...
    libTKernel.so.7.8 => /home/user/.local/lib/libTKernel.so.7.8
    libTKMath.so.7.8 => /home/user/.local/lib/libTKMath.so.7.8
    libTKG3d.so.7.8 => /home/user/.local/lib/libTKG3d.so.7.8
        libTKG2d.so.7.8 => /home/user/.local/lib/libTKG2d.so.7.8
    libTKGeomBase.so.7.8 => /home/user/.local/lib/libTKGeomBase.so.7.8
[...]

PR #767, opened 2024-03-11
Adding support for OpenCascade 7.8.0

thierry-FreeBSD:
This patch allows to build with OpenCascade 7.8.0.
It resolves https://github.com/horizon-eda/horizon/issues/764 .

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -31,7 +31,11 @@ endif
 if not cxx.links('int main(void) {return 0;}', dependencies: opencascade)
     message('opencascade link args are broken, replacing')
     # on ubuntu 20.04, the link args are broken and are missing the -l for some reason??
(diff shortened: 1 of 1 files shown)


carrotIndustries:
Just wondering, why are the link args broken in the first place so that it has to take that code path?

aperezdc:
> Just wondering, why are the link args broken in the first place so that it has to take that code path?

OpenCascade 7.8 has reorganized the library names, it's mentioned in the [upgrade notes for 7.8](https://dev.opencascade.org/doc/overview/html/occt__upgrade.html#upgrade_780_tk_rework).

carrotIndustries:
> OpenCascade 7.8 has reorganized the library names, it's mentioned in the [upgrade notes for 7.8](https://dev.opencascade.org/doc/overview/html/occt__upgrade.html#upgrade_780_tk_rework).

Still, this code path should only be taken if opencasacade was packaged in a broken way to begin with.

carrotIndustries:
Should be fixed by https://github.com/horizon-eda/horizon/commit/afa25d7e50f75ced34047cf8c3ffd17e3090cffe

PR #777, opened 2024-06-06
ensure XDG_CACHE_HOME exists

dl6tom:
while packaging Horizon EDA for OpenBSD i noticed, that when launched from a freshly created user account it fails with
```
libc++abi: terminating due to uncaught exception of type horizon::SQLite::Error: unable to open database file
Abort trap (core dumped)
```
that happens cause XDG_CACHE_HOME  (`~/.cache` in this case) does not exist in the home directory yet.

this change ensures it exists and also temporarily changes umask to ensure permissions according to XDG Base Directory Specification.

### DIFF
--- a/src/util/util.cpp
+++ b/src/util/util.cpp
@@ -2,7 +2,9 @@
 #include <fstream>
 #include <giomm.h>
 #include <glibmm/miscutils.h>
+#include <glib/gfileutils.h>
 #include <glib/gstdio.h>
+#include <sys/stat.h>
 #include <unistd.h>
 #ifdef G_OS_WIN32
 #include <windows.h>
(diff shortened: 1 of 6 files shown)


src/util/util.cpp
```diff
@@ -215,11 +216,20 @@ std::string get_config_dir()
     return Glib::build_filename(Glib::get_user_config_dir(), "horizon");
 }
 
-void create_config_dir()
+void create_cache_and_config_dir()
 {
+    // XDG base directories should be created as 0700
+    auto previous_umask = umask(S_IRWXG | S_IRWXO);
```
carrotIndustries, line 222: Isn't there a better way that doesn't involve global state?

dl6tom: [Gio::File::make_directory_with_parents|https://gnome.pages.gitlab.gnome.org/glibmm/classGio_1_1File.html#afb0b5b094339057b4b15952eb73e08d6] has no way to specify permissions and it seems glibs [g_mkdir_with_parents|https://docs.gtk.org/glib/func.mkdir_with_parents.html] has no equivalent in glibmm. i could use the C function from glib of course.

Issue #744, opened 2023-09-24
Wrong usage of Board Editor CLI leads to SIGABRT instead of error message


alex-fu27:
I ran into this here: #743
The usage of the Board Editor is `horizon-imp -b <board file> <block file> <via directory>` and if one does not specify all of the files, this leads to:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 1) >= this->size() (which is 1)
```
from `src/core/core_board.hpp:15`.

It would be nice to have an error message here that gives the correct usage.

alex-fu27:
Super weird, since I wrote this issue, it started telling me `wrong invocation` :confused:

carrotIndustries:
I'm happy to merge a PR for this, but this isn't a top priority for me since it's only used for debugging and development.

PR #780, opened 2024-06-21
Add messages to help with horizon-imp cli arg usage for #744

frmdstryr:
Hopefully this is correct. Let me know if there's a better way to do it. I made it just give a warning instead of exit.

### DIFF
--- a/src/imp/imp_main.cpp
+++ b/src/imp/imp_main.cpp
@@ -41,7 +41,7 @@ static void expect_dir_arg(const std::vector<std::string> &filenames, const char
 
 static void expect_pool_env(const std::string &pool_base_path)
 {
-    if (pool_base_path.length() <= 0) {
+    if (pool_base_path.empty()) {
         g_warning("HORIZON_POOL dir env var is not set");
     }
     else if (!Glib::file_test(pool_base_path, Glib::FILE_TEST_EXISTS)) {
--- a/src/imp/imp_main.cpp
+++ b/src/imp/imp_main.cpp
@@ -67,7 +67,7 @@ int main(int argc, char *argv[])
 
     Glib::OptionContext options;
     options.set_summary("horizon interactive manipulator");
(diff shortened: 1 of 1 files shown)


src/imp/imp_main.cpp
```diff
@@ -25,6 +25,27 @@
 
 using json = nlohmann::json;
 
+#define EXPECT_FILENAME_ARG(name, i)                                                                                   \
```
carrotIndustries, line 28: why do these have to be macros?

src/imp/imp_main.cpp
```diff
@@ -25,6 +25,33 @@
...
+static void expect_dir_arg(const std::vector<std::string> &filenames, const char *name, size_t i)
+{
+    if (filenames.size() <= i || filenames.at(i).size() == 0) {
+        g_warning("%s directory argument at %i missing", name, static_cast<int>(i));
+    }
+}
+
+static void expect_pool_env(const std::string &pool_base_path)
+{
+    if (pool_base_path.length() <= 0) {
```
servusdei2018, line 44: `pool_base_path.empty()` would be more idiomatic

src/imp/imp_main.cpp
```diff
@@ -123,41 +151,59 @@ int main(int argc, char *argv[])
...
     else if (mode_package) {
+        expect_filename_arg(filenames, "package", 0);
+        expect_pool_env(pool_base_path);
         imp.reset(new horizon::ImpPackage(filenames.at(0), pool_base_path, temp_mode));
         if (filenames.size() > 1)
             imp->set_suggested_filename(filenames.at(1));
     }
     else if (mode_board) {
+        expect_filename_arg(filenames, "board", 0);
+        expect_filename_arg(filenames, "planes", 1);
```
carrotIndustries, line 184: The planes file is somewhat optional. If it's not there, it's not worth a warning that may confuse people.

frmdstryr: Thanks. It'll ignore the file exists check for that argument now.

PR #784, opened 2024-09-18
Allow packages to fix parameters

jmwilson:
This change allows packages and pads to define fixed parameters, which cannot be overwritten by a board's parameter rule. This can be useful in a number of situations, such as a package that wants to define a courtyard excess based on IPC density level, or has specific solder and paste mask settings that are recommended by the manufacturer. While these results can also be obtained by manually defining the courtyard boundary or a adding a shape on the mask layer, parametric definitions are much nicer.

This change bumps the pool version to 1, and adds `parameters_fixed` as a field in packages and pads. When the board pushes its parameter rule to packages via `Package::apply_parameter_set`, the parameter ID is checked against the set of fixed parameters. When both a pad and package fix the same parameter, the pad's value will take precedence.

In the parameter set editor UI, the option to fix a parameter is an additional button that appears with the pin icon, which is similarly used in many other programs. When multiple pads are selected in the parameter editor, using the `All` button will copy the fixed field to all pads, and clicking the `All` button on a fixed parameter will also copy it to all pads.

While fixed parameters could conceptually apply to padstacks, padstacks also allow required parameters that must be defined by a pad. Having a parameter that is both fixed (at the padstack level) and required of pads using it would be somewhat contradictory, so padstacks cannot define fixed parameters.

### DIFF
--- a/src/dialogs/pad_parameter_set_window.cpp
+++ b/src/dialogs/pad_parameter_set_window.cpp
@@ -44,7 +44,11 @@ PadParameterSetWindow::PadParameterSetWindow(Gtk::Window *parent, class ImpInter
         for (auto &it : pads) {
             it->pool_padstack = ps;
             it->padstack = *ps;
-            it->padstack.apply_parameter_set(it->parameter_set);
+            auto ps_this = it->parameter_set;
+            copy_param(ps_this, pkg.parameter_set, it->parameters_fixed,
+                       {ParameterID::SOLDER_MASK_EXPANSION, ParameterID::PASTE_MASK_CONTRACTION,
+                        ParameterID::HOLE_SOLDER_MASK_EXPANSION});
+            it->padstack.apply_parameter_set(ps_this);
         }
         emit_event(ToolDataWindow::Event::UPDATE);
     });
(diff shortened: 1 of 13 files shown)


src/pool/package.cpp
```diff
@@ -118,7 +118,7 @@ int Package::Model::get_rotation(unsigned int ax) const
     }
 }
 
-static const unsigned int app_version = 0;
+static const unsigned int app_version = 1;
```
carrotIndustries, line 121: Reflect this in https://github.com/horizon-eda/horizon/blob/master/scripts/app_versions.yml

src/imp/imp_package.cpp
```diff
@@ -169,6 +169,14 @@ void ImpPackage::check_alt_pkg()
     pool->db.execute("DROP VIEW IF EXISTS pkg_deps");
 }
 
+class MyToggleButton : public Gtk::ToggleButton, public Changeable {
```
carrotIndustries, line 172: Why this extra class?

src/dialogs/pad_parameter_set_window.cpp
```diff
@@ -79,14 +99,52 @@ void PadParameterSetWindow::load_pad()
         delete editor;
     editor = Gtk::manage(new ParameterSetEditor(&pad_current->parameter_set, false));
     editor->set_has_apply_all_toggle("Apply to all selected pads (Shift+Enter)");
+    editor->signal_create_extra_widget().connect([this](ParameterID id) {
+        auto w = Gtk::manage(new Gtk::ToggleButton);
```
carrotIndustries, line 103: I think a "Fixed" checkbox is more intuitive than the pushpin toggle button.

src/dialogs/pad_parameter_set_window.cpp
```diff
@@ -12,6 +12,22 @@
 
 namespace horizon {
 
+static void copy_param(ParameterSet &dest, const ParameterSet &src, const std::set<ParameterID> &parameters_fixed,
```
carrotIndustries, line 15: These look to be identical to the ones in package.cpp. So maybe make them static in the `Package` class.

src/imp/imp_package.cpp
carrotIndustries: Yeah, had a look at that code again and that class is also unnecessary there. Must have been some leftover. Removed it in 49704012a845b7fbf539f7370c6c9b08f56f7584.

jmwilson: Wait, without that, how does changing the toggle or checkbox signal that the state has changed and the package/padstack needs to be saved? Before I added this, I would have to change one of the numerical properties for the editor to see the fixed parameter change and ask to save the changes. What does the padstack editor do differently?

src/dialogs/pad_parameter_set_window.cpp
jmwilson: I think having them as functions in `src/parameter/set.cpp` is best

src/imp/imp_package.cpp
jmwilson: Oh, I see a way to make it work, by binding the editor object and making it emit `signal_changed`:

    editor->signal_create_extra_widget().connect([this, editor](ParameterID id) {
        auto w = Gtk::manage(new Gtk::CheckButton("Fixed"));
        w->set_tooltip_text("Fixed parameters cannot be changed by board rules");
        w->set_active(core_package.parameters_fixed.count(id));
        w->signal_toggled().connect([this, id, w, editor] {
            if (w->get_active()) {
                core_package.parameters_fixed.insert(id);
            }
            else {
                core_package.parameters_fixed.erase(id);
            }
            editor->signal_changed().emit();
        });
        return w;
    });

The line `editor->signal_changed().emit();` is also needed in `imp_padstack.cpp` after 49704012a845b7fbf539f7370c6c9b08f56f7584 removed the signaling from `MyCheckButton`

carrotIndustries: Thanks. that's indeed needed. However, the commit where I removed `MyCheckButton` didn't make the situation worse since nothing connected to its `signal_changed` and there's no mechanism that'd automatically propagate the `signal_changed` up through the widget hierarchy.

src/dialogs/pad_parameter_set_window.cpp
carrotIndustries: Sound even better, go ahead.

jmwilson:
Thanks for the comments! I added dynamic versioning and switched to a checkbox in the parameter windows: e80720c6e552b475788731379da05d5e84464e02

Screenshot 2024-09-27 081457
Screenshot 2024-09-27 081415

src/imp/imp_package.cpp
jmwilson: I think the link was here since `MyCheckButton` inherited from `Changeable`: https://github.com/horizon-eda/horizon/blob/b979731880c390a91e9d5fb182b96a21010632a8/src/widgets/parameter_set_editor.cpp#L67-L72

Does it make sense to remove that snippet now too?

carrotIndustries: > I think the link was here since MyCheckButton inherited from Changeable: 

Oh you're right. that's some spooky action-at-a-distance I added back then.

As far a as I can tell that code path isn't taken anymore so we can remove the check for `Changeable` there.

jmwilson:
@carrotIndustries any more thoughts on this?

carrotIndustries:
LGTM, just squash the two last commits, then I'll merge it

carrotIndustries:
Done, thanks for the high-quality PR!

PR #787, opened 2024-10-10
Prevent duplicate 3D model files

jmwilson:
The model map doesn't check for duplicate values, so you can add the same file several times, which seems like a bug. 

Also, for packages that use multiple 3D models, it seems convenient to allow selecting multiple files in the chooser. Any chosen files that are already present get ignored.

### DIFF
--- a/src/imp/3d/imp_package_3d.cpp
+++ b/src/imp/3d/imp_package_3d.cpp
@@ -42,7 +42,49 @@ void ImpPackage::update_points()
     }
 }
 
-std::string ImpPackage::ask_3d_model_filename(const std::string &current_filename)
+std::vector<std::string> ImpPackage::ask_3d_model_filenames()
+{
+    GtkFileChooserNative *native = gtk_file_chooser_native_new("Open", GTK_WINDOW(view_3d_window->gobj()),
+                                                               GTK_FILE_CHOOSER_ACTION_OPEN, "_Open", "_Cancel");
+    auto chooser = Glib::wrap(GTK_FILE_CHOOSER(native));
+    auto filter = Gtk::FileFilter::create();
+    filter->set_name("STEP models");
+    filter->add_pattern("*.step");
+    filter->add_pattern("*.STEP");
+    filter->add_pattern("*.stp");
+    filter->add_pattern("*.STP");
+    chooser->add_filter(filter);
+    chooser->set_select_multiple(true);
+    chooser->set_current_folder(pool->get_base_path());
+
+    while (1) {
+        if (gtk_native_dialog_run(GTK_NATIVE_DIALOG(native)) == GTK_RESPONSE_ACCEPT) {
+            auto base_path = Gio::File::create_for_path(pool->get_base_path());
+            std::vector<std::string> rel_names;
+
+            for (const auto &file : chooser->get_files()) {
+                std::string rel_path = base_path->get_relative_path(file);
+                replace_backslash(rel_path);
+                rel_names.push_back(rel_path);
+            }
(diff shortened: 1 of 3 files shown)


src/imp/3d/imp_package_3d.cpp
```diff
@@ -55,7 +97,8 @@ std::string ImpPackage::ask_3d_model_filename(const std::string &current_filenam
     filter->add_pattern("*.STP");
     chooser->add_filter(filter);
     if (current_filename.size()) {
-        chooser->set_filename(Glib::build_filename(pool->get_base_path(), current_filename));
```
jmwilson, line 58: https://docs.gtk.org/gtk3/method.FileChooser.set_filename.html

says this should only be used in a save chooser, I found it caused problems with the directory defaulting to the last used directory instead of the model directory

src/imp/3d/model_editor.cpp
```diff
@@ -83,16 +83,21 @@ ModelEditor::ModelEditor(ImpPackage &iimp, const UUID &iuu)
 
         auto browse_button = Gtk::manage(new Gtk::Button("Browse…"));
         browse_button->signal_clicked().connect([this, entry] {
-            Package::Model *model2 = nullptr;
-            if (imp.core_package.models.count(uu)) {
```
jmwilson, line 87: I'm unable to a see a code path where `imp.core_package.models[uu]` goes away between the constructor and here, so I removed the check and directly use `imp.core_package.models.at(uu)`. If the delete button is used then this button goes away too.

src/imp/3d/imp_package_3d.cpp
carrotIndustries: Interesting, for me it does work.

src/imp/3d/imp_package_3d.cpp
```diff
@@ -42,7 +42,49 @@ void ImpPackage::update_points()
...
+            auto base_path = Gio::File::create_for_path(pool->get_base_path());
+            std::vector<std::string> rel_names;
+
+            for (const auto &file : chooser->get_files()) {
+                std::string rel_path = base_path->get_relative_path(file);
+                replace_backslash(rel_path);
+                rel_names.push_back(rel_path);
+            }
+            if (std::none_of(rel_names.cbegin(), rel_names.cend(),
+                             std::bind(std::string::empty, std::placeholders::_1))) {
```
carrotIndustries, line 71: This fails to build on all platforms and compiler versions in the CI. As far as i can tell what you're trying to do here is undefined behavior since you're trying to take the address of a standard library function: https://en.cppreference.com/w/cpp/language/extending_std

I suggest using a lambda function instead.

jmwilson: @carrotIndustries looks like the CI is happy after switching to a lambda

carrotIndustries:
LGTM, i'd like to keep ` chooser->set_filename` for now as this is somewhat separate from this improvement/bugfix.

jmwilson:
Hmm, I suspect it's a GTK issue with `set_filename` behaving broken on Windows but fine on Linux, since that's where I saw this behavior. If you want to strip it out, I'll do it but it seems like it gives better cross-platform behavior.

carrotIndustries:
To me at least the file picker defaulting to what's in the entry is part of a quality user experience.

If it doesn't work on windows, I think the best way around this is an `#ifdef G_OS_WIN32` to get the best thing possible on all platforms.

carrotIndustries:
I think this a useful feature to have, but should be separated from the `set_filename` issue that can be discussed and fixed independently.

PR #789, opened 2024-10-29
Highlight board editor unplaced package list selection in schematic

frmdstryr:
When placing components on the board it currently only highlights the component in the schematic after clicking place. This makes it highlight the component when it is selected in the unplaced components list.

It'd also be nice to have the reverse, eg make selecting in the schematic highlight the entry in the unplaced list but I didn't add it.

[Screencast_20241029_102249.webm](https://github.com/user-attachments/assets/96c6566f-3689-49e4-8a9b-8709713ba184)

### DIFF
--- a/src/imp/imp_board.cpp
+++ b/src/imp/imp_board.cpp
@@ -791,6 +791,18 @@ void ImpBoard::construct()
         }
         this->tool_begin(ToolID::MAP_PACKAGE, true, components);
     });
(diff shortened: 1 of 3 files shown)


carrotIndustries:
LGTM, rebase onto lastest master to fix the CI.

frmdstryr:
LMK if anything else should be changed.

PR #811, opened 2025-06-30
unvendor nlohmann-json and range-v3

monego:
nlohmann-json and range-v3 are common packages and are available in Meson's WrapDB.

The commits delete the vendored library, add a check in a meson.build and replace include quotes with brackets.

Unvendoring these saves ~3MB in repository size.


carrotIndustries:
So this will then download the dependencies at build time? I'm not really a friend of that approach and many downstream packaging systems don't have network access at build time. This means we'd then need to provide release tarballs that include these dependencies.

Also, I don't feel easy merging a PR with more changes than what's possible to review. So provide the script that you used to replace quotes with angle brackets.

Another approach I'd be more inclined to merge is using range-v3/nlohmann-json from the system package manager.

So overall, don't expect this PR to merged in its current form.

monego:
> So this will then download the dependencies at build time? I'm not really a friend of that approach and many downstream packaging systems don't have network access at build time. This means we'd then need to provide release tarballs that include these dependencies.
> 
> Also, I don't feel easy merging a PR with more changes than what's possible to review. So provide the script that you used to replace quotes with angle brackets.
> 
> Another approach I'd be more inclined to merge is using range-v3/nlohmann-json from the system package manager.
> 
> So overall, don't expect this PR to merged in its current form.

The idea is to use the the downstream system packages, the same way other dependencies outside of 3rd_party are handled. I only mentioned WrapDB because of Windows, but honestly I don't know how dependencies are managed there.

For the replacement, I used fd and sed (from the repository root):

```bash
fdfind . src --type f --exec sed -i 's|#include "nlohmann/json.hpp"|#include <nlohmann/json.hpp>|g' {}
fdfind . src --type f --exec sed -i 's|#include "nlohmann/json_fwd.hpp"|#include <nlohmann/json_fwd.hpp>|g' {}
```

No automation was done in range-v3, there are few files including it and they had angle brackets already.

carrotIndustries:
For this project there only should be two kinds of dependencies:
 - bundled into the `3rd_party` directory
 - provided by the system package manager

Right now, the oldest somewhat supported distro we build on is ubuntu focal (20.04), though we could bump this to 22.04 if it helps.

If we find that a dependency is provided by a all major distros in a sufficiently recent version, we can remove the bundled dependency and rely on the system package manager.

What would work in either case is using meson to detect that the system package manager provides a dependency in the appropriate version and then configure the include paths and built projects to use the system-provided dependency instead.

PR #833, opened 2025-12-07
Fix opensuse CI build

fk0815:
Switching install to devel_python3 pattern uses latest python version.

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -153,7 +153,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Install dependencies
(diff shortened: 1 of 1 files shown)


carrotIndustries:
Latest master works though, https://github.com/horizon-eda/horizon/actions/runs/19719366673 re ran-the opensuse builds today

fk0815:
ok, then my "fix" does not fix the problem. I found

'python3' not found in package names. Trying capabilities.

which is due to no python3 package exists, but there is also

'pkg-config' not found in package names. Trying capabilities.

which is also resolved automatically. After again checking the failed case I think the fix should be to add

 --allow-downgrade

to the zypper in command.

PR #834, opened 2026-01-05
Check for mistyped/copied ".pdf" ext in datasheet url

frmdstryr:
I had copy / pasted a url and accidentally skipped the last character causing the link to not work.  

Since most datasheets are pdfs, this shows a check warning if the datasheet url ends with "." or if it has ".p" but doesn't match ".pdf". 

image


carrotIndustries:
Seems reasonable, but fix the windows build. Use u8string in there somewhere.

Issue #742, opened 2023-09-19
Feature request: add an option to make all 3D views use orthographic projection by default


torque:
For CAD applications, I find orthographic projection to be much more useful because it makes misalignment and/or misdimensioning problems very visually obvious. Every mechanical CAD program I am aware of uses orthographic projection by default in its design views.

I've found myself toggling the projection type in the 3D view whenever I open it, so it would be nice to have a global option I could set.

carrotIndustries:
There are several ways how to we could implement this:

 - Add a setting in the preferences window
 - Automatically remember the last setting as a default

PR #836, opened 2026-01-10
Add default 3d view preferences for #742

frmdstryr:
This adds a "3D View" section to the "Editor" preferences tab with new settings to set the default view projection and colors.

image

[Screencast_20260109_211710.webm](https://github.com/user-attachments/assets/296381f4-87fb-4983-a5d7-3e4ca5e7118d)


carrotIndustries:
How does this interact with the solder mask, silkscreen and substrate colors stored in the board?

PR #853, opened 2026-07-20
Clear net diffpair on both nets

frmdstryr:
IDK if it's intentional or if something broke but currently the clear diff pair tool appears to remove one net from the pair and leave the other (at least visually on the schematic).

https://github.com/user-attachments/assets/8f85cf34-57f8-454b-b71a-7c1f40fcab8a


frmdstryr:
That works better. Thanks!

PR #854, opened 2026-08-27
UX: Don't duplicate alternate names when creating new unit pin

griush:
So basically, in the Unit Editor, when you put alternate names in a pin and then create a new pin, it duplicates the alternate names the pin might have. As it may seem obvious, it's very rare that pins have the same alternate names and I found myself removing them by hand very often. With this line removed, alternate pins are basically empty when creating a new pin.


carrotIndustries:
Thanks for the PR. PRs like this one are always nice to see to confirm that people are in fact using Horizon EDA ;)

Issue #256, opened 2019-05-07
Feature Request: command line interface


niclashoyer:
Provide a command line interface to automate creation of BOM, gerber export or DRC.

I tried to automate KiCad (which has no real CLI) and resorted to a docker container running xvfb which is a PITA and breaks continuously.

carrotIndustries:
I've though about this as well and it's certainly within the realms of possibility. One reason why we have no CLI yet is not having a clear idea of what it should be like:

 - Provide a limited set of features such as check drc, export pdf, etc accessible using command line args
 - Launch some sort of script interpreter that you'd tell to open the schematic and export the pdf (TCL would be the classical choice)
 - Create a python module (implemented in C/C++, calling into the horizon code) to the same as above:

From a programmer's perspective, having something based on python to achieve the latter would be preferable, but that doesn't package very well on windows.

Just writing from the top of my head of how that python thing could look like
```
>>> import horizon
>>> brd = horizon.load_board("path/to/project.hprj")
>>> settings = brd.get_fab_output_settings() #returns the fab output settings serialized to a dict
>>> #do something with the settings
>>> brd.export_gerber(settings)
```

Since windows users probably aren't interested in that kind of feature at all, the packaging issues are acceptable from my perspective.

niclashoyer:
If the python interface is simple to use, that would be the best idea. But if you take KiCad as an example, which also has a python interface for the board designer, it gets complicated quickly, even for [simple tasks as checking DRC](https://forum.kicad.info/t/kicad-drc-for-travis/13852/4). There are even blogs just about [kicad scripting](https://kicad.mmccoo.com/). There are a lot of python scripts out there for KiCad, but most of them are unmaintained, don't do what the actually should do, or reinvent the wheel. Of course one can do fancy stuff with that (e.g. generative board outlines), but for simple tasks this might be too much.

Maybe a limited CLI and a scripting interface is orthogonal, so it may be good to have both. From an CI perspective you'd just want to check the rules, convert schematic to pdf, export BOM, convert the board(s) to gerber/pdf.

carrotIndustries:
> If the python interface is simple to use, that would be the best idea.
> Maybe a limited CLI and a scripting interface is orthogonal

I'm not planning on exposing all of horizon's guts in a python module. That module will just provide features similar to the ones you mentioned.

carrotIndustries:
[...]

PR #855, opened 2026-09-09
 Add command-line project exports for CI and automation

atoav:
Hi @carrotIndustries  et. al.,

this one has been on the back of my mind for a while.

It adds an `export` command to the as of now mostly under-utilized CLI of the `horizon-eda` binary. This enables the export of project artifacts like schematic PDFs, gerbers and BOMs from shell scripts and CI jobs without starting the GUI or connecting to a display server. 

I know in prior discussion (https://github.com/horizon-eda/horizon/issues/256) the solution of choice mentioned by you was python. I feel that adding this is would still be an improvement in terms of comfort and not require users to build the horizon.so from source.

### Motivation

The motivation is to let projects using Horizon-EDA to simply regenerate their documentation, fabrication files and assembly data from a checked-out revision, for example when building a release. I have found this a bit tedious and error-prone myself, since it is easy to forget a single export step, especially if I just made some minor changes. This way it is much easier to ensure the exported files are always in sync with a repos Horizon-Project files.

### General Overview

The added commands reuse the existing exporters and the project's saved export settings and can be used as follows:

```sh
horizon-eda export schematic project.hprj -o artifacts/schematic.pdf
horizon-eda export board project.hprj -o artifacts/board.pdf
horizon-eda export gerber project.hprj --output-dir artifacts/gerbers
horizon-eda export bom project.hprj -o artifacts/bom.csv
horizon-eda export pnp project.hprj --output-dir artifacts/assembly
horizon-eda export step project.hprj -o artifacts/board.step
horizon-eda export odb project.hprj --output-dir artifacts/odb
```

I also took the freedom to add a general`horizon-eda --version` command as I find this is somewhat useful for build logs. I hope this is acceptable.

Schematic export includes instantiated child blocks, Gerber export includes Excellon drills and ODB++ supports the saved directory or archive format. The `schematic` and `board` commands use the output extension to select the format, just like in the UI currently only `.pdf` is supported, but if for example SVG export would be added at some point, users could just type `-o board.svg` and then this could be used to select the new format.

### Output path setting

Each command requires an explicit output path. This was a deliberate choice, because the GUI allows relative paths and I found it ambiguous to decide from which "perspective" to interpret the paths. Normally one would expect a CLI tool to express paths relative to the cwd, but the stored paths in the settings are relative to the project directory. So I decided to make it always relative to the CLI and ignore stored paths. 

### Setting Overrides

The export commands accept a `--settings FILE` for JSON overrides. This allows users to store multiple settings and export multiple variants. E.g. I have always found it useful to export the board pdf for assembly both for top and bottom views of the board. However the export settings in the UI only remember one of the settings. This way I can create two json files and have a script do the magic for me, without changing the program settings.

### Other Options

Because it is rude to delete user files there is an`--overwrite` option, so users cannot accidentally delete existing files. A `--quiet` option does exactly what it says on the tin. 

### Technical Details

Diagnostics go to stderr, with exit code 1 for loading or export failures and 2 for invalid arguments or settings.

The CLI builds a temporary pool index from the project's local pool and cached items, so jobs do not need personal pool configuration or a committed `pool.db`. It leaves project inputs unchanged, recalculates copper planes for the relevant exports, and treats loading warnings as failures to avoid publishing incomplete data. Output is staged before publication, with checks for conflicting filenames and destinations that would replace project inputs.

The dispatcher runs before GTK initialization and leaves other invocations to the existing application entry point. The implementation lives in `src/cli`, with separate modules for arguments, settings, project pools, exporter adapters and output handling. I hope the way I did it is okay, I have little experience with this type of C++ project since my C++ experience is mostly with embedded.

I moved the schematic expansion and block flattening into `src/project/export_util.cpp`, so the CLI and Python bindings can use the same code. Some of the existing writers also needed to report output failures, otherwise a script would think the export succeeded even though writing the files failed. For STEP export I added an optional "strict mode", so missing or unreadable models and invalid board outlines fail the CLI export. Existing callers keep the default behavior.

### Documentation

I also updated the manual page and added `doc/cli.md` with the commands, settings examples and some notes on how the code is organized. Hopefully this makes it easier both to use the exports in a script and to add another exporter later on. Maybe this should go to the docs page instead tho.

### Tests

For testing, I added a small generated project so the tests do not depend on somebody's local pool. It covers all seven headless exporters, including child blocks, images, populated and unpopulated components, copper planes, drills and placement on both sides of the board. The tests also check the different ODB++ formats, failing STEP exports, settings overrides and output conflicts, as well as making sure the project files are left alone. Both suites pass with `meson test -C build cli tests --print-errorlogs`.

### Some points about CI

There are a couple of things to keep in mind when using this in CI. The referenced pool items and any enabled STEP models still need to be available locally. Also, while the files are generated in a temporary directory first, copying several files to the output directory is not an all-or-nothing operation. If that step fails, some files may already have been copied, so a CI job should only publish its artifacts after the export command has succeeded.

### Disclosure on LLM usage

I am not sure what the policy of this project on LLM usage is, so I disclose to which degree LLMs have been involved: 

I used LLMs to
1.  find my way into the codebase
2. check my code against potential problems I may have overlooked
3. to suggest solutions and explain code in one or two places where I got stuck because my understanding of modern C++ has still room to improve
4. I used an LLM to generate the test code (`tests/cli_fixture.cpp` and `cli.py` 

I did *not* use LLMs to:
- generate any comment, commit message, documentation or pull request message (= this one) for human consumption (I find this incredibly disrespectful)
- generate blindly used code 

I read every line of code added via this PR, some of them many times.


atoav:
Okay I investigated the failing CI jobs a little bit and these appear to be two different issues that have existed in commit 1c19f96549140aeb842fe1c27889642375fa555f already, so they are unrelated to my PR.

## FreeBSD

```
error: unknown type name 'TopTools_ListOfShape'
``` 
in `export_step.cpp` (line 454)

appears to be related to OpenCASCADE, there are quite a number of deprecation notices, for example:

```
Standard_HEADER_DEPRECATED("TDF_LabelSequence.hxx is deprecated since OCCT 8.0.0. Use ...
```

To me this indicates that the FreeBSD-Install installs a newer OpenCASCADE that is newer than 8.0.0, so maybe something changed in the interface binding there 


## Windows

Here we see:
- `PdfGraphicsStateWrapper has no SetFillColor or SetCurrentMatrix`
- `Matrix::FromCoefficients no longer exists`
- etc.

Which also points to an API incompatibility, I presume with PoDoFo, since that is the PDF dependency right?

carrotIndustries:
> failing CI jobs

Yeah, podofo released another API-incompatible version yet again…
So far only freebsd ships opencascade 8.0. Need to setup a freebsd VM or throw some cents a VPS to get a dev environment to make it compatible with 8.0.

Overall, all this would be better off in its own binary that doesn't depend on gtk at all rather than bolting this onto `horizon-eda`.
Option parsing should be done using the Glib option parser, see imp_main.cpp.

I don't quite see the point of that `ExportPool` thing, the project pool should contain everything regardless of locally-installed pools.

To make reviewing the PR easier, put general improvements in a separate PR.

src/export_bom/export_bom.cpp
carrotIndustries, line 1: separate PR

src/export_gerber/excellon_writer.cpp
carrotIndustries, line 1: separate PR

src/export_gerber/gerber_writer.cpp
carrotIndustries, line 1: separate PR

src/export_pnp/export_pnp.cpp
carrotIndustries, line 1: separate PR

src/export_step/export_step.cpp
```diff
@@ -559,7 +563,7 @@ void export_step(const std::string &filename, const Board &brd, class IPool &poo
         hdr.SetOriginatingSystem(new TCollection_HAsciiString("horizon EDA"));
         hdr.SetDescriptionValue(1, new TCollection_HAsciiString("Electronic assembly"));
 
-        if (Standard_False == writer.Write(filename.c_str()))
+        if (writer.Write(filename.c_str()) != IFSelect_RetDone)
```
carrotIndustries, line 566: separate PR

src/python_module/schematic.cpp
carrotIndustries, line 1: separate PR

src/project/export_util.cpp
carrotIndustries, line 1: separate PR

src/python_module/board.cpp
carrotIndustries, line 1: separate PR

atoav:
Thank you for taking the time to wrap your head around this. Hope you're doing fine : )

> Overall, all this would be better off in its own binary that doesn't depend on gtk at all rather than bolting this onto `horizon-eda`. Option parsing should be done using the Glib option parser, see imp_main.cpp.

That is quite a good idea. Does `horizon-cli` make sense? I will of course use the Glib option parser if that is what is already being used elsewhere.


> I don't quite see the point of that `ExportPool` thing, the project pool should contain everything regardless of locally-installed pools.

Then I maybe misunderstood the whole thing, I will try to replace this with the Project Pool.

> To make reviewing the PR easier, put general improvements in a separate PR.

Okay I will try to do that, thanks again for taking the time.

atoav:
Okay I now broke out the changes as follows:

- #857 adds a more descriptive error return status to the exporters, as this is needed for errors on the CLI
- #858 refactors the exporter code so both the Python module and the cli tool can rely on the same code
- #859 moves the CLI code out into its own `horizon-cli` binary and implements your suggested changes (option parsing, using project pool) as well as the documentation changes needed. This PR obviously depends on the changes  in #857 and #858

The last PR is intended to replace the changes in this one, so my suggested course of action would be:

1. #857 and #858 could be merged first if you accept the changes. 
2. Once merged, I rebase #859 onto master so its diff only contains the remaining CLI changes
3. Once #859 is ready to take over, I will close this PR in favor of it

What do you think?

PR #861, opened 2026-09-14
improvement: now main branch name is also looked at in the pool editor

griush:
My main motivation here is that git changed the default branch name to `main` instead of `master` meaning that new pools like mine that use `main` as default branch name basically cannot use the git tab in the pool editor.

Thanks for making this project.


carrotIndustries:
LGTM, fix the code style by running `git diff origin/master --name-only  | grep '\.cpp\|\.hpp' | xargs clang-format -style=file -i`

src/pool-prj-mgr/pool-mgr/pool_git_box.cpp
```diff
@@ -682,7 +683,13 @@ void PoolGitBox::handle_back_to_master(bool delete_pr)
     }
 
     autofree_ptr<git_object> treeish(git_object_free);
-    git_revparse_single(&treeish.ptr, repo, "master");
```
carrotIndustries, line 685: At that point, introduce a helper function like `find_main_branch` that's basically `git_revparse_single(&obj.ptr, repo, "master") != 0 && git_revparse_single(&obj.ptr, repo, "main") != 0` in one place

griush:
I did a simple static function and not a class method just because seemed simpler.

Issue #797, opened 2025-01-19
Feature Request: Support for Castellation PCB


oldrev:
Hello, I recently noticed that KiCAD has added support for creating castellation PCB, and I really hope Horizon can offer the same feature. 

I know that the edge-placed vias in Horizon can be recognized as castellated holes by the fab, but it would be great to have an option in the Padstack to show the castellated effect in the 3D preview. 

Horizon does need some more love, Thank you.

carrotIndustries:
So is it just a thing in the 3D preview? Started hacking something and it now looks like this: 

![Image](https://github.com/user-attachments/assets/4e64dcc6-da6b-4287-9506-2fe415944e5c)

Obviously still need to make the pth barrel open at the board edge.

oldrev:
Great job! IMHO, the best approach would be to add a "Castellation" option to the "Type" dropdown in the Padstack Editor, so it won’t mess up with the DRC.

![Image](https://github.com/user-attachments/assets/f18ff3f2-8082-4feb-be9e-b8285f3f997b)

oldrev:
So, is the current 2.6 version safe enough to give it a try with castellated holes?

![Image](https://github.com/user-attachments/assets/68d01b23-4e41-4665-9db5-09ec04fc3791)

carrotIndustries:
Nice board!

I've made at least one board with castellated holes and it turned out as expected.

What ultimately matters are the gerbers and in there, having the hole cut by outline layer is the expected way of representing castellated holes. See https://jlcpcb.com/help/article/what-is-castellated-holes for example.

Also make sure to specify castellated holes when ordering your board since some PCB manufacturers charge extra for them.

oldrev:
Thanks for the reminder, I've done a few castellated boards with KiCad before. 

This board is just a fun project anyway，it's okay to fail. I'm not gonna spend $$$ on antenna tuning or FCC/CE certifications for it.

PR #862, opened 2026-09-18
3D: support castellated board edge vias

oldrev:
This PR fixes the 3D rendering issue when using THT Pads and Vias as half-holes on the PCB edge #797. To minimize changes, clipping of THT Pads and Vias at the board outline edge was added only for CanvasMesh. Hope it helps.

The result:
image


src/canvas3d/canvas_mesh.cpp
```diff
@@ -449,6 +484,77 @@ void CanvasMesh::prepare_layer(int layer)
     }
 }
 
+static ClipperLib::IntRect get_path_bbox(const ClipperLib::Path &path)
+{
+    ClipperLib::IntRect bbox{path.front().X, path.front().Y, path.front().X, path.front().Y};
```
carrotIndustries: use `BBoxAccumulator` from `util`

src/canvas3d/canvas_mesh.cpp
```diff
@@ -260,6 +267,18 @@ void CanvasMesh::prepare(const Board &brd)
...
+
+    board_outline.clear();
+    const auto it = ca.get_patches().find({PatchType::OTHER, BoardLayers::L_OUTLINE, UUID()});
+    if (it != ca.get_patches().end()) {
+        ClipperLib::Clipper cl;
+        cl.AddPaths(it->second, ClipperLib::ptSubject, true);
+        cl.Execute(ClipperLib::ctUnion, board_outline, ClipperLib::pftEvenOdd);
+        if (!board_outline.empty()) {
+            board_outline_bbox = cl.GetBounds();
+            std::swap(board_outline_bbox.top, board_outline_bbox.bottom);
```
carrotIndustries: since Y axis of clipper is flipped?

src/canvas3d/canvas_mesh.cpp
```diff
@@ -415,7 +434,23 @@ void CanvasMesh::prepare_layer(int layer)
         for (const auto &it : ca.get_patches()) {
             const auto l = layer_is_substrate(layer) ? BoardLayers::L_OUTLINE : layer;
             if (it.first.layer == l) {
-                cl.AddPaths(it.second, ClipperLib::ptSubject, true);
+                const bool clip_copper = BoardLayers::is_copper(layer)
```
carrotIndustries, line 444: ah, this one takes care of the the pads?

src/canvas3d/canvas_mesh.cpp
```diff
@@ -449,6 +484,77 @@ void CanvasMesh::prepare_layer(int layer)
...
+    cl.AddPath(open_path, ClipperLib::ptSubject, false);
+    cl.AddPaths(board_outline, ClipperLib::ptClip, true);
+    cl.Execute(ClipperLib::ctIntersection, tree, ClipperLib::pftNonZero, ClipperLib::pftNonZero);
+    ClipperLib::OpenPathsFromPolyTree(tree, arcs);
+    auto &walls = layers.at(layer).walls;
+    for (const auto &arc : arcs) {
+        if (arc.size() < 2) {
+            continue;
+        }
+        walls.emplace_back(2 * arc.front().X - arc[1].X, 2 * arc.front().Y - arc[1].Y);
```
carrotIndustries, line 547: what's going on with the coordinate transformations here?

src/canvas3d/canvas_mesh.cpp
```diff
@@ -459,7 +565,11 @@ void CanvasMesh::add_path(int layer, const ClipperLib::Path &path)
         }
         layers.at(layer).walls.emplace_back(path[0].X, path[0].Y);
         layers.at(layer).walls.emplace_back(path[1].X, path[1].Y);
-        layers.at(layer).walls.emplace_back(NAN, NAN);
+        auto &walls = layers.at(layer).walls;
+        while ((walls.size() + 1) % 4 != 0) {
```
carrotIndustries, line 566: what's the purpose of `walls.size() + 1) % 4 != 0`?

carrotIndustries:
LGTM overall, just some questions to understand the code should it ever require some maintenance

src/canvas3d/canvas_mesh.cpp
oldrev: Yeah, I'm using BBoxAccumulator now, so it's no longer needed.

src/canvas3d/canvas_mesh.cpp
oldrev: Ah, here, I'm not very familiar with the rendering part of the code. 

I thought the vertices needed to stay a multiple of 4, but it turns out they don't. I've already removed it, and the test results look fine too.

src/canvas3d/canvas_mesh.cpp
oldrev: Yes.

src/canvas3d/canvas_mesh.cpp
oldrev: It adds endpoint neighbors to each open arc so the renderer can determine its direction at both ends.

PR #864, opened 2026-09-20
Fix/pool database corruption

benefrasch:
Sometimes when updating the pool, there was an error that the database disk image was malformed.
The main cause was the overflow that is fixed with this PR.
Some other hardening changes included as well.


3rd_party/alphanum/alphanum.hpp
```diff
@@ -117,39 +117,31 @@ namespace doj
...
-
-	      // get the right number
-	      unsigned long r_int=0;
+		++l;
+	      const char *r_start = r;
 	      while(*r && alphanum_isdigit(*r))
+		++r;
+
+	      // Compare digit runs without converting them to an integer. Numeric
+	      // conversion overflows for long values and breaks transitivity,
```
carrotIndustries: where exactly do we run into this? the collation using this function is only used for part MPN and manufacturer.

src/pool-update/pool-update.cpp
```diff
@@ -21,6 +22,13 @@ void pool_update(const std::string &pool_base_path, pool_update_cb_t status_cb,
 {
     if (!status_cb)
         status_cb = &status_cb_nop;
+
+    // Keep updates serialized across threads and processes. The transaction is
+    // released automatically when this connection is closed, including errors.
+    SQLite::Database update_lock(Glib::build_filename(pool_base_path, ".pool-update.lock"),
```
carrotIndustries: pool updates only happen in  `horizon-eda` which is a single-instance application, so this can be handled in-process

3rd_party/alphanum/alphanum.hpp
carrotIndustries: I think this may indeed the be main culprit. With MPNs such as `74404064220` that are larger than a 32 bit int. On 64 bit windows, a long is 32 bit, whereas on 64 bit linux, it's 64 bit which could explain why we're running into this.

Instead of this vibe-coded patch of questionable provenance, I'd prefer merely bumping the integers to `uint64_t`

benefrasch: Also fine by me if that solves it.
However, I did encounter this error on Linux as well which disappeared with this patch.
Whether you want to use the Copilot vibepatch (whether it works or not) or create your own fix, I don't mind.

PR #865, opened 2026-09-22
fix height restriction deletion branch

oldrev:
So I'm trying to add a copper thieving plane feature to Horizon, and while I was goofing around, I noticed that the deletion in `ToolAddHeightRestriction::begin()` seems to have a bug, it's not deleting the `height_restriction`, it deleting the `keepout` by mistake.


carrotIndustries:
Yeah, now you know where the height restriction tool comes from ;)

